PR #457, opened 2015-03-08
Added Vietnamese translation

thanhtai2009:


pbatard:
Thanks you very much. I will apply it shortly.

Issue #483, opened 2015-04-04
Add a tooltip for the language selection button


RonS2015:
That globe icon doesn't really make it clear what it's for. A tooltip would be nice.

pbatard:
I think that's a fair request.

I'll see what I can do, but this will require rallying 30 translators to translate the tooltip, so it will probably be a couple of months at best to get this enhancement added.

Halamix2:
Or change the globe to flag?

pbatard:
Rufus is very size conscious, so wherever it can, it uses icons from the Windows system rather than one embedded (Why? Because embedded icons have would have to scale nicely for high DPI displays, and a nicely scaling icon usually adds 100KB!).

IMO, the current globe + arrow icon is actually better at conveying that the button allows language selection than any of the other icons that are native to a Windows system. But I'm open to suggestions for a better icon if you can find one (NB: you can use the excellent [IconsExtractor from NirSoft](http://www.nirsoft.net/utils/iconsext.html) to poke system DLLs, such as `shell32.dll`, for icons)

Also, if I were to pick a flag, which flag should I pick? As pointed out, I cannot embed all 30+ flags for the languages Rufus supports, as it would probably increase the size of the app by half. And using only one flag would be confusing IMO...

pbatard:
Done in 91e699d38a5cb2ee66a9ef3d8c69d0d2cff40bb8.
Translator updates will be added when I finalize the list of new translations needed (#504), which will happen __after__ 2.3 is out.

Issue #243, opened 2013-12-07
Keyboard accessibility


dandunckelman:
First off, this is a sweet lil tool.

I just wanted to say that I just used this to make a Windows 7 bootable USB from an ISO on a Windows Vista machine.  Unfortunately, this particular machine had a faulty keyboard and touchpad...so I don't have any input source unless I plug in a usb keyboard/mouse.

Being that I love using the keyboard exclusively, I thought I'd be cool and only plug in a keyboard.  So, I open the open, but as soon as I open it, I can't use the keyboard to do anything (except maybe starting the process with all default settings).

I hope this makes sense.  Definitely, this is a nice to have feature and not a big requirement.  I just had to plug in a mouse I had lying around to interact with the app's menus.

Thanks again and happy weekending!

pbatard:
Adding the following issue when using Jaws, reported today by __Aman Singer__:

> If I click the box to select UEFI as the partition type, I can move with the arrows among the choices, which is good, but then I have to move the mouse around and if that mouse cursor lands on the previous choice, the box switches to the choice the mouse is on.

pbatard:
Adding the following issue when using nvda screen reader, reported by __Mike Coulombe__:
> When I open the program I get no speech telling me which box i'm on when I press tab or arrow keys. However, pressing space does open the current menu and I can then use tab to make the choices because I get speech. Would you be able to get the main menu to also work with nvda?

PR #488, opened 2015-04-18
[ui] change language button

Quppa:
Language button is now a toolbar button with a dropdown arrow. Icon is a globe for Windows XP to Windows 7 and a globe with characters on it for Windows 8 and above (same as the icon for the 'Languages' control panel).

The button should have a tooltip, but adding this will involve updating all translations (see issue #483).


pbatard:
Wow, that looks really great. And you covered everything in your code, including RTL. Nice job! :+1: 

I'll apply this patch as soon as I get a chance (there's a very minor issue with WDK compilation, but I'll fix that). And I'll make sure the tooltip gets added.

Thank you so much for submitting this improvement!

Quppa:
Thanks for the positive feedback :)

For adding a tooltip, it looks like setting the [TBSTYLE_EX_MIXEDBUTTONS extended style](https://msdn.microsoft.com/en-us/library/windows/desktop/bb760430%28v=vs.85%29.aspx) makes it easy.

I just pushed a minor change that means the button is aligned to the right edge of the 'device list' combobox instead of 12 (device independent) pixels from the right edge of the window. Setting the margin to 12 DIPs manually for some reason meant that the combobox and button were 1 pixel out of alignment at 125% DPI (but not at 100% or 150% DPI).

Not related to the language button, but do you avoid including the 256x256px icon to reduce the executable size (despite that icon existing in res/icon-set)?

pbatard:
> I just pushed a minor change that means the button is aligned to the right edge of the 'device list' combobox instead of 12 (device independent) pixels from the right edge of the window.

I've now picked that addon and will be pushing the full commit shortly.

I think I made sure that the Device List was always 12 DIP from the right edge, and the alignment looked OK for me at 200% DPI, but I didn't test 150%. I'm actually planning to overhaul some of the positioning in the near/middle-term future because Rufus looks doesn't look too great at high DPI. For instance, the image selection button is taller than the combobox, and the space between combo / checkboxes is weird. I now realize that I should have done what [putty](http://www.chiark.greenend.org.uk/~sgtatham/putty/download.html) does, which is ditch `RC` specified controls altogether and instead create each control programmatically (in the same way you created and positionned the language toolbar). The more I am dealing with `RC` controls, the more I realize that nobody in their right mind should ever want to use that...

> Not related to the language button, but do you avoid including the 256x256px icon to reduce the executable size (despite that icon existing in res/icon-set)?

Yup. Adding the 256x256px image to the iconset results in a size increase of more than 100KB, which is way too much. The problem with icons is that Microsoft uses a very stupid format, that can't be compressed (well, __officially__ the 256x256px icon __can__ be compressed, but the compression algorithm Microsoft picked is really poor, and I was barely able to gain 5% when I tried -- for reference,  LZMA will easily save more than 50%) because it needs to be accessible without running the app.

Maybe when 4K screens become the norm, I'll add the 256x256px image, but for now, I'd rather add useful code and features rather than eye-candy, while keeping the executable size small(ish).

This being said, I will be adding the 64x64px image to the icon for v2.2 (largest icon in current set is only 48x48px), to at least get better looking toolbar and About box icons at 200% DPI. This will add about 15KB to the exe though :(


Now, with regards to your patch, I modified a few minor things, including not calling `ImageList_AddIcon()` and `ImageList_Create()` directly, as this create an issue when producing XP compatible apps with MinGW. I also changed a few variable names as well as some formatting, but nothing major. I also dropped setting the language toolbar as a tabstop, as it would get selected by default when launching the app, which isn't what I want. But I'll review the tabstop situation when I look into keyboard accessibility (#243). And of course, I'll look into adding the tooltip when I deal with #483.

All in all, I must really thank you for spending time on this - I makes the Rufus UI much nicer!

pbatard:
NB: Actual commit is 3579c770a212e74ab9560116c8b334062f0a16db not 2f167c965f96fe2aebb816a7ad15023722cbfd52.

Quppa:
RC files can be a pain, but managing everything yourself isn't necessarily a light undertaking. I would have liked to position the toolbar in the RC file, but there's no TOOLBAR type :(

Fair point about the icon size. Unfortunately only the 256x256px image in the icon gets PNG compressed - in the case of rufus.ico, the 128x128px image ends up taking up more space than the 256x256px image! I guess you won't be rushing to add support for Windows 10's new 768x768px icon size :)

Thanks for fixing up those few things in my commits. Regarding issue #243, I too noticed that the tab key doesn't work in the dialog - I made another PR that improves keyboard navigation (#491).

PR #491, opened 2015-04-19
[ui] make the tab and arrow keys work in the dialog

Quppa:
The tab and arrow keys can now be used to navigate around the various dialogs.


pbatard:
I'm starting to feel like it's Christmas here, coz this is another awesome gift from you... :gift:

You've really made my day here. And with keyboard navigation working as it should at long last, you're gonna make a lot of other people happy!

The only thing I changed from your patch is defaulting to the `Select ISO` button instead of `Close` when there's no USB, because I think this is what most people will want (and what Rufus was doing before your patch). I also added an `hStart` variable to replace all these calls to `GetDlgItem(hDlg, IDC_START)` because we have more than enough of them to justify the use of a global.

This should greatly improve things for #243 (but I'm not closing that one for now, as I still need to check what the various screen readers for Visibility Impaired people will do).

Can't thank you enough for spending time on this! :smile:

Quppa:
You're welcome. All the controls already had the `WS_TABSTOP` style, so most of the work was already done :)

Good idea to make a global `hStart`.

Two quite minor things I only noticed after committing:
- I forgot to delete the comment on line 1707.
- `IDC_ABOUT_COPYRIGHTS` (the second RichEdit control on the About dialog) should probably be a tab stop, as currently there's no way to scroll it without using the mouse.

PR #599, opened 2015-09-22
Updated greek translation to 1.0.18

nmargaritis:


pbatard:
Thanks. I will apply this right after I'm done with Danish (which is being delayed due to some clarifications needed).

PR #624, opened 2015-10-22
Update Traditional Chinese translation

Chocobo1:
* I added a commit "Trim trailing spaces", since my text editor always do that, it becomes a bit annoying to re-add the spaces myself, hope you could merge it.
* There is an online localization platform called "Transifex". I think it might help the current lack of translator situation.


pbatard:
Applied. Many thanks for this! :smile:

Issue #490, opened 2015-04-18
New volume label will return to ISO image label after i finished a quick format only


Sopor-:
If i load an image file the 'New volume label' will change to the label i have on the image file. If i instead unchecked all except 'Quick format' i still have the label from the image file. I then change it because i want another name on my USB-device. I then start the quick format and when it's done it now change back to the label on the image file. So i was thinking, have it used my label or have it used the label from the image file?
When i put in a USB-device i can see the label on the device and when i load an image file it will change to that label but i suggest that it should only do that if the 'Create a bootable disk using' is selected and when i only quick format a device i have it unchecked so when the format is finished it should show the label of the USB-device instead of the image label.

Short version: Even if the image file is loaded it should only show the label from the image file if 'Create a bootable disk using' is selected. If i unselect it, show the USB-device label instead.

pbatard:
I'll see what I can do, but since this is a minor UI __preference__, and UI changes are always a major pain in the ass to handle (because changing one minor visible feature, usually impacts a whole bunch of user behaviour paths that are already coded in the app) I will keep this a __super-low__ priority.

PR #667, opened 2016-01-09
Fixed issue #490

Int80x:
Issue #490

Basically just checked to see if the bootable checkbox was checked when populating UI, if it isn't it just sets the volume label textbox to the current volume label.

### DIFF
--- a/src/rufus.c
+++ b/src/rufus.c
@@ -687,7 +687,11 @@ static BOOL PopulateProperties(int ComboIndex)
 			SetWindowTextU(hLabel, DriveLabel.String[ComboIndex]);
 		}
 	} else {
-		SetWindowTextU(hLabel, img_report.label);
+		if (IsChecked(IDC_BOOT)) {
+			SetWindowTextU(hLabel, img_report.label);
+		} else {
+			SetWindowTextU(hLabel, DriveLabel.String[ComboIndex]);
+		}
 	}
 
 	return TRUE;


pbatard:
Sorry, but I don't think this patch addresses the issue raised by @Sopor, who, if I understand correctly, wants the proposed label to go back to the existing USB label if the user unchecks the _Create bootable disk using_ check box.

This means that `PopulateProperties()` is not the right place to do it, as it doesn't get called on check/uncheck event for that checkbox.

The place to change the label would be in the `case IDC_BOOT:` section, and you'd have to test whether the checkbox is being checked or unchecked to see what label needs to be set. You also have to be careful to do it in the same fashion as is done in `PopulateProperties()`, so you'd probably want to factorize the label setting code into a new function call.

For the time being, I'm afraid I'll have to reject this patch. But thanks for trying to help!

Int80x:
I think it does, read it one more time, his English isn't that great.

pbatard:
Well, I think there are 2 elements to address in #490:
1. Label should not be set to the currently selected ISO image after format, if _Create bootable disk using_ is not selected
2. Label should swicth between currently selected ISO image and existing USB label, when checking/unchecking _Create bootable disk using_. At least that's how I interpret _"If i unselect it, show the USB-device label instead" (and I think it makes a lot of sense to have that too).

I did test the proposed patch and I'm afraid that, while it does address part 1, it is missing part 2.
That's why I don't think I want to apply a patch that does only half the job, especially when it's possible that even if we apply a patch for part 1, part 2 might require a rework of it. 
If possible, I'd rather have a patch that handles both parts.

Int80x:
That's fair. I'll work on the second part tonight.

pbatard:
Sounds good - thanks for that!

PR #669, opened 2016-01-10
Fixed issue #490 and added drag-n-drop support

Int80x:
I'm really new to this and don't exactly know how to make separate pull requests, so I just deleted my old one and made a new one.

Fixed the issue that you had with my last pull request, adding "part 2".

Also, I really appreciate the advice you gave me about development, so here's some drag-n-drop support :)

You can now just drag a .iso file anywhere onto the main window and it will act as if you selected one from the file browser.

I don't know why it says I deleted 492 lines of code (or added 528), because I didn't, but whatever.

### DIFF
--- a/src/rufus.c
+++ b/src/rufus.c
@@ -687,7 +687,11 @@ static BOOL PopulateProperties(int ComboIndex)
 			SetWindowTextU(hLabel, DriveLabel.String[ComboIndex]);
 		}
 	} else {
-		SetWindowTextU(hLabel, img_report.label);
+		if (IsChecked(IDC_BOOT)) {
+			SetWindowTextU(hLabel, img_report.label);
+		} else {
+			SetWindowTextU(hLabel, DriveLabel.String[ComboIndex]);
+		}
 	}
 
 	return TRUE;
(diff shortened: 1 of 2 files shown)


pbatard:
Okay, I have applied your changes for #490, but as you will see, I had to heavily modify and expand on them (See 4b3969e475d5b6ecb5b4459777277b89c1a36739) to fully address what I think needed to be addressed with that issue, besides the 2 parts I already mention, which are:
1. Ensure that there is a consistent policy that is always used to apply the labels
2. Ensure that when a user selects `FreeDOS` in the dropdown (or something else than ISO), the label reverts to the existing one, instead of the one from the currently selected ISO
3. Ensure that this also fits with the existing `user_changed_label` boolean, that was used to try to keep a user input label where possible (and which is more tricky than it looks, as I found that `EN_CHANGE` messages are also generated when the app modifies the label, so I had to add an extra `app_changed_label` boolean.

At this stage, I'm still not sure I haven't broken something with setting up the USB label... and I'm got the feeling `user_changed_label` was broken in the first place anyway. But this patch is really what I feel was needed to fully address the issue reported by @Sopor. And I hope it gives you greater understanding as to why I will sometime defer __seamingly__ trivial issues, because they rarely get addressed with only 4 lines of new code... :cry: 

Still, I appreciate the help with the patch. I will close this pull request now, even as know the drag and drop proposal has not been reviewed, as I will kindly request that you submit a new version that doesn't modify styling/whitespaces or parts that shouldn't be modified for review.

PR #751, opened 2016-05-10
Update zh-TW Translation

Chocobo1:


pbatard:
Many thanks for the quick patch!

PR #762, opened 2016-05-27
Update Turkish loc

Tmp341:
Updated to the latest.


Tmp341:
Btw, i didn't do that deletion.

pbatard:
I'm sorry, but I have to reject this request.

I appreciate that you're following the changes to Rufus, but __please__ hold on from submitting change requests for translations until __AFTER__ I have updated [`res/localization/ChangeLog.txt`](https://github.com/pbatard/rufus/blob/master/res/localization/ChangeLog.txt) with a new version.

In other words, until you see changes for `1.0.21` being introduced in `ChangeLog.txt`, I will not be accepting translation updates, even if the English version has changed.

The reason for that is that there may be many more messages being added/modified for the next version, and with 35 translators to go after, if I were to add every single translation changes as they are introduced, instead of grouping them shortly before a release, I'd spend my life dealing with translation updates.

So, please wait until you see [`res/localization/ChangeLog.txt`](https://github.com/pbatard/rufus/blob/master/res/localization/ChangeLog.txt) being updated. Thank you.

Tmp341:
I thought you'll keep it open until you think it's addable.

pbatard:
Nah. I don't want to keep pull requests open for months, and I think I should be able to remember it.

PR #801, opened 2016-07-24
Use modern FAT32 boot records

pluto00987:
Hi.

This adds checks and actions to use nt52 or nt60 FAT32 boot records when appropriate.

The nt60 (ie PE) BR is used extensively for Win PE 2.0+ boot disks.

The nt52 (ie NT) BR may or may not be used in anything common but it might as well be available if the case comes up.


pbatard:
Thanks, but I'm not sure I see the point of having that in Rufus, especially as we don't create anything `ntldr` or `bootmgr` bootable using FAT32.

As you will have seen from the source, we deliberately chose to ignore the `fat_32_pe` part of ms-sys because there is no real need for it: the only (BIOS) bootable FAT32 drives Rufus creates are for Linux or DOS (as well a few oddities like KolibriOS or ReactOS), but not Windows. Anything BIOS/Windows is forced to NTFS, so that, in most cases, it can't be booted from UEFI, which avoids people installing in UEFI mode when they wanted BIOS mode, and vice versa. So this is the very __intentional__ reason why there aren't any calls for `write_fat_32_pe_br()` or `write_fat_32_nt_br()` in the first place. In other words, the new boot record writing cases you are adding with this patch are not meant to ever be selectable. 

So the only thing this may add is the ability to report, when you plug a drive, whether it has an existing fat32pe or fat32nt boot record, which is about one line in the log...

Do you have an actual usage scenario, besides detection of the boot record written by a tool other than Rufus, for this patch?

Depending on the increase in size, I may still decide to add the detection part (so that the ms-sys part from Rufus is closer to the current ms-sys), but I don't think I'll add the boot record writing part.

pluto00987:
Well, primarily the use case I mentioned: Win PE 2.0+. While I realize Rufus defaults to NTFS for such images, one is still able to click the drop-down and switch back to FAT32. Rufus will happily create a drive with those settings, but as-is it results in a non-working boot disk (BIOS at least). In addition, the cause isn't the least bit obvious. Even as a very experienced system admin, it took me several hours of futzing and testing to finally figure out that Rufus was writing an old-school DOS BR when FAT32 is selected.

I had already read the FAQ about dual-BIOS/UEFI booting being purposefully disabled so I did have a hunch it was 'broken' on purpose. I mostly wrote the patch for fun. I guess it was wishful thinking that you'd accept it.

Still, sticking with your intention, I would suggest either actually disallowing the selection of FAT32 (or FAT or exFAT), or popping up a warning saying THIS WON'T DO WHAT YOU WANT or something to that effect. That might save some future users some hair-pulling like I dealt with.

pbatard:
> one is still able to click the drop-down and switch back to FAT32. Rufus will happily create a drive with those settings

I don't think you are reading the information that is presented to you correctly then.

If you use a Windows ISO and try to select FAT32 with "MBR for BIOS or UEFI-CSM", then Rufus will flat out tell you that this is not supported. That is the case even if you try to use the <kbd>Alt</kbd>-<kbd>E</kbd> trick.

So, really, the __ONLY__ way to get Rufus to accept FAT32 with a Windows ISO is if you select "GPT for UEFI" or "MBR for UEFI", in which case, Rufus quite explicitly indicates that, whatever you do, __the drive
will NOT be BIOS bootable__.

> THIS WON'T DO WHAT YOU WANT

Sorry, but if someone did change the __default__ "MBR for BIOS or UEFI-CSM" to "GPT for UEFI" or "MBR for UEFI", which they need to do if they select a Windows ISO (as it will always default to the firts one), then they should realize that they have just removed all promise of booting from BIOS.

An additional message would pretty much equate saying "Hey, user, you don't really read what Rufus is telling you, are you? What part of being only UEFI bootable when only UEFI is present in the dropdown is so hard to understand?", which I'd rather not do for obvious reasons.

pluto00987:
> If you use a Windows ISO and try to select FAT32 with "MBR for BIOS or UEFI-CSM", then Rufus will flat out tell you that this is not supported. That is the case even if you try to use the Alt-E trick.

I'm sorry, but that's not the case. Using Rufus 2.10, I select my PE 4.0 ISO, and the settings are populated with "MBR for BIOS or UEFI-CSM"and "NTFS". I can then change the file system to "FAT32" and the partition scheme DOES NOT CHANGE. Then I click 'start' and Rufus will go about its business without a complaint.

Maybe what you say is _supposed_ to happen, but it does not. Perhaps some iso check in there isn't accurate for newer PE images.

pbatard:
Then I'll need to fix the check, as I only tested it with __official__ Microsoft images, and of course, PE 4.0 ISOs should behave the same. Do you have an example of a public PE 4.0 based ISO I could download to test?

pbatard:
Additionally, a full log fro, Rufus of how you created your FAT32 with Rufus, from a PE 4.0 image, would help.

Issue #807, opened 2016-08-05
Proposal: GUI improvements


SetTrend:
### The Rufus GUI is confusing in regard to user feedback and automatic argument updates.

So I'd like to suggest the following GUI improvements:

- [ ] Add a read-only field displaying the currently selected ISO file path.
- [ ] Users read the GUI from top to bottom. Selecting a new ISO file alters a previously edited drive label and partitioning scheme. So I suggest to put the ISO file selection line _above_ these input elements.
- [ ] Switch from a one-line action log to a multi-line action log with vertical scroll bar. (Containing just the information from the one-line log, not the detailed log appearing after clicking the `Log` button.)

<br/>
### To sum up, the GUI should change from this:
![rufus1](https://cloud.githubusercontent.com/assets/9283914/17425668/6d1bc1e8-5ad0-11e6-8b43-eedbcd39d391.png)

... to this:
![rufus3](https://cloud.githubusercontent.com/assets/9283914/17425944/362712b6-5ad3-11e6-8259-98523f0e1719.png)

pbatard:
This is pretty much a duplicate of #117. __HOWEVER__, proper UI changes (that scale correctly at multiple DPIs rather that the current [atrocious hit-and-miss that the default Windows APIs provide](https://github.com/pbatard/rufus/blob/master/src/rufus.c#L1098-L1100) and which is what Rufus currently uses) are __MUCH MORE__ difficult to implement and will require __A LOT OF TIME__ to get through. So that's why the UI enhancements have been on hold for couple of years now.

They will happen. But it might very well be another year before they do.

> Add a read-only field displaying the currently selected ISO file path.

I'm already planning to do something like this (but\not\with\the\full\path\which\could\potentially\make\the\thing\way\too\long\and\which\99.99%\of\users\should\remember). Also, you can always get the currently selected ISO from the log, and it usually remains displayed for a while on the status bar. But yes, this is already something I am planning to improve.

> So I suggest to put the ISO file selection line above these input elements.

Yup. That's been planned for a long time. It's an improvement that's kind of hard to miss.

> Switch from a one-line action log to a multi-line action log with vertical scroll bar. 

[...]

PR #808, opened 2016-08-05
GUI amended

SetTrend:
* Read-only field added showing selected ISO file.
* Input elements reorganized to reflect workflow
* Revised one-column layout to become two-column layout, having the main window fit 800*600 pixels.

rufus


SetTrend:
I have updated the English and German localization only. If you accept the changes I'll try to revise other languages, too.

pbatard:
As I explained with #807, I have to reject this request for the time being. Sorry.

SetTrend:
Er ... all work is done ... You just have to merge. No work needed from your side.

pbatard:
Let me be clearer. Your proposed changes is NOT how I see the UI for Rufus 3.0, which is the version of Rufus where I will be redesigning it. As the creator of Rufus, I hope I am entitled to follow my vision for the UI, which I have been thinking about for the past couple of years, rather than have to accept someone else's. Just moving stuff around is only a small part of how I want the redesign to happen.

I do appreciate your effort, but I just can't see myself applying the changes you propose, so, thanks, but no thanks.

SetTrend:
Sure, no problem.

I had just been wondering why it should take a couple of years to amend the GUI when it can be done in a few hours. I was supposing you'd probably be involved in more important tasks.

After all, I was just trying to help.

BTW: From what I can tell, you've got a good coding style. It was very easy to survey and straightforward to amend. That's a cachet which is not very common nowadays. :clap:

pbatard:
> I had just been wondering why it should take a couple of years to amend the GUI when it can be done in a few hours. I was supposing you'd probably be involved in more important tasks.

Since you spent some time on this, I'm going to try to answer that.

One of the first thing I want from the redesign is something that will look good at various DPIs and follow the Windows UI design guidelines. For a start, we're not using the font size that Microsoft recommends for modern UI design, so that needs to be fixed. But if we do that, then we need to review all the move/resize business from the .loc file. And then there's the whole "We had to compute an empirical formula that makes no sense to find out how much space we need to add at various DPIs when we move stuff around (e.g. advanced mode)". Furthermore, Microsoft is really screwing up UI developers with RC's (but IMO forms is worse coz it adds A LOT of completely superfluous overhead), and we also had to apply ugly workarounds to keep XP and non XP users happy in terms of fonts (and even then, Microsoft still managed to stab us in the back with their braindead use of Unicode). So a __MUCH__ better solution then would be to design something that does away with placing elements in an RC file, so that things flow without having to reposition/resize them manually, and so that, if we decide to adjust font size (or, if has happened in the past, Microsoft decides with Windows 11 that UI should use a different font/font size), we won't have to repeat this whole excruciating exercise of positioning and sizing stuff around, for the base RC as well as for the 36 translations.

So that means writing a mini UI framework that will properly size and position the various UI components. This is pretty much an independent component (such as [bled](https://github.com/pbatard/bled/), which I designed when I needed compression support, and found that none of the existing libs satisfied my needs, or [efifs](https://github.com/pbatard/efifs), of which I only needed the NTFS driver, but with which I decided to support as many file systems as poissible, in case it could help other people). So before I can even start moving a single element for Rufus 3.0, there'll probably need to be a 'noRC' project, that takes care of what I described above in a semi-generic fashion, and that I can integrate in Rufus so that I __never ever have to go through a costly manual move/resize exercise when I design a new win32 application or redesign an existing one___. That's a long term goal, i.e. something that needs to be crafted properly and with enough dedicated time, to avoid something half assed that falls apart at the first real hurdle. My current estimate is that doing this properly will take about __6 months__, during which I don't want to have to be distracted with another major feature.

So, why hasn't this been started yet even though the UI redesign request has been opened for about 3 years now?

Well, for a start, UI redesign is always lower priority compared to non existing features. Let me make that a bit clearer with an example: If people want Windows To Go support, and it's not existing in Rufus, they'll probably be much happier if they get it sooner (even if they have to use a somewhat quirky UI to access the feature) rather than later. So what happened between the time I knew I wanted to redesign the UI and now is me concentrating on adding features (some of which major, necessitating many months of work) rather than rework existing ones like the UI.

Now, because I know that many people are growing impatient about this, I was aiming at starting the UI redesign/noRC work in January this year. However, due to various circumstances, some of which were major a, that couldn't happen. For instance, I still decided prioritizing adding non USB card reader support (which had been requested by many Rufus users) and speeding up checksum computations (when your tool is downloaded A LOT, you kind of have an obligation to save the millions of people who use it as much time as you can, as, the way I see it, not doing so globally adds up to months or years being collectively wasted waiting in front of a screen!), as well as other stuff (you can always take a look at the commit history or ChangeLog to see what I focused on)

SetTrend:
We're absolutely on the same page. Microsoft RC dialog design and its quirks is hell to work with. I've been struggling with that fact for years, too. (And at the eve of Windows 8.1 they finally noticed themselves, too. Hard to believe they acted deaf for decades.)

May I humbly suggest to switch to [Microsoft WPF](https://msdn.microsoft.com/en-us/library/aa970268(v=vs.100).aspx) for your GUI design. It is Microsoft's solution to all your concerns and encapsulates localization, fonts and automatic dialog resizing using a grid layout approach. WPF encapsulates dn deals with all the gory details/peculiarities of the different operating systems.

pbatard:
Yeah, I looked into WPF (which is what I called "forms" above), but I'm not convinced. It adds way too much overhead IMO, so I'd rather see what can be achieved without, and if we go full fledged framework, then I'd rather go Qt, so that it could ease the pains of porting Rufus to Linux or OS X (which I'm not really planning to do, but which someone else might) as far as UI is concerned. Maybe I'll change my mind once I've started my own UI mini-framework, but for the time being, I don't think I should have to go the full UI framework route, just as I didn't go the gettext route for localization, but created my own thing which I think turned quite okay.

And of course, that's not to say I won't take suggestions in terms of redesign. Maybe if more people complain about the UI size, I'll look into dual column (or, which is the solution I'd prefer, if I can get a good "reflow" from my mini framework, let users decide that for themselves). But I need to have that UI framework first, just like I wasn't ready to accept translations until I had support for `.loc` files.

I can only apologize for the time I know you spent on the redesign &mdash; it's simply still too early for me to accept these requests.

PR #850, opened 2016-11-10
Fix case sensitiviness issues by using always lowercase

aperezdc:
This makes it possible to cross build in a machine with a case-sensitive file system. Tested in Arch Linux, where the following commands will successfully complete a build:

```sh
% sudo pacman -S mingw-w64  # Only if it wasn't already installed
```

```
% sh configure --host=i686-w64-mingw32 && make -j$(nproc --all)
% file src/rufus.exe
src/rufus.exe: PE32 executable (GUI) Intel 80386, for MS Windows
```

Doing 64-bit builds seems to be working as well:

```
% sh configure --host=x86_64-w64-mingw32 && make -j$(nproc --all)
% file src/rufus.exe
src/rufus.exe: PE32+ executable (GUI) x86-64, for MS Windows
```


pbatard:
Thanks, but the only DOS file that actually needs to be renamed is `MODE.COM`.

I tend to keep the case of the files as they are supplied by the FreeDOS project, which is why the `EGA`'s are lowercase, and the rest aren't. So I'd rather not move everything to lowercase when it isn't needed, as next time I update FreeDOS, we may end up with case alteration.

For now, I just renamed `mode.com` &rarr; `MODE.COM` (since that follows the case I have in both the `.rc` and [here](https://github.com/pbatard/rufus/blob/master/src/dos.c#L354-L359)). Of course, I also picked the `Windows.h` to `windows.h` (which is something Visual Studio automatically capitalizes on its own, and which I always fight again... but it looks like I had missed this one).

While I was at it, I also applied the `x` permission to the script files, which, at the very least, should allow to invoke `./configure`.

Please test and let me know if you still see an issue.

aperezdc:
@pbatard: Your explanation about keeping the capitalization in FreeDOS makes sense :wink: 

I have tried running a build locally with MingGW in Linux after your changes, and it works perfect. Thanks!

pbatard:
Glad to hear that. Thanks again for the patch!

PR #905, opened 2017-02-17
[loc] fix a typo in Simplified Chinese translation

buaawjw:


pbatard:
Thank you very much!!

PR #974, opened 2017-06-28
Fix ortographic errors

trinaldi:
An user - Ricardo Pereira da Costa - contacted me and suggest/show me some fixes.


pbatard:
Applied. Thank you very much!

trinaldi:
Pete, let me ask you.

When someone suggests a change, should I edit right away or wait for some
kind of review?

Thanks

pbatard:
Hi Tiago, sorry for the late reply. How and when you submit it is pretty much up to you, as I'm not really in a position to review changes for languages that I don't speak. Generally speaking, I prefer if a series of changes for a specific translation are grouped into a single commit, but it's not a big deal if they aren't. But it's really up to you to decide what you think is best. Thanks.

NickeManarin:
I speak Portuguese and I agree with this changes.

PR #975, opened 2017-06-29
Update rufus.loc

marcellogianola:


pbatard:
Sorry for the late update - I now have applied your changes to the translation, thank you very much for this!

PR #978, opened 2017-07-05
fix memory leak in do_convert() in src/libcdio/driver/utf8.c

j123123:


pbatard:
Thanks, but can you send this patch to the libcdio mailing list instead?

Rufus is reusing the code from libcdio here, and as much as possible I try to avoid deviating too much from libcdio upstream, so that I can more easily pick up updates from this library.

If you get the libcdio team to fix this, then that fix will naturally find its way into Rufus, as I tend to update the libcdio code on regular basis.

For the record, there are quite a few potential issues similar to this one, that I am well aware of, with the libcdio code, but that I am not planning to address in Rufus as I'd rather have them addressed upstream. You can get a good idea for that if you ran the libcdio part through Coverity or other static code analysis tools.

Ever since I started using libcdio and Coverity in Rufus, I've been meaning to produce patches for those, and send them to the libcdio people for review and integration, but I never managed to prioritize this over other things, as most of these issues don't seem that problematic (a memory leak because realloc doesn't have enough memory for the new buffer is usually the least of one's worries). Still, I sure wouldn't mind if people could send patches to the libcdio project to address them.

For now, I'm afraid that I will just close that request, again, not because the patch is unneeded, but because it should be a part of libcdio first.

PR #982, opened 2017-07-18
[loc] update Chinese Simplified translation to latest

buaawjw:


pbatard:
Applied. Many thanks for this!

PR #1012, opened 2017-08-29
Merge pull request #1 from pbatard/master

mfloresn90:
Pull request


pbatard:
There's no code change. If you want to propose a pull request, can you please make sure there's actually something valid to pull?

NSbuilder:
Well, I'm quite sure that's a mistake, it's not like someone will want to commit dummy pull requests anyway.

pbatard:
The problem is that it's the second time I'm getting an empty pull request this week... I hope this is not the beginning of a trend, especially as I would expect it's hard to create a pull request by mistake.

PR #1019, opened 2017-09-09
Update rufus.loc Lithuanian translation

embar-:


pbatard:
Sorry but I can't accept this translation update, as it messes up the UI:
image1
As you can see `Paleidimo laikmeną kurti naudojant` is truncated on high DPI displays, and to correct this, the size of the dialog as well as most of the combo boxes need to be adjusted..

Also, I would very much prefer if you contacted the current Lithuanian translator (whose e-mail is provided in the About dialog) to discuss your proposed changes with them. As long as there is an active translator, and especially if the changes are not trivial, I don't want to update translation behind the back of the current translator.

So please get in touch with the current translator, and make sure you both agree with a proposal, as well as ensure that the UI elements don't get truncated.

embar-:
Now few more changes, including changing „Paleidimo laikmeną kurti naudojant“ to shorter „Sistema paleidžiama iš laikmenos“

embar-:
I wrote letter to Gintaras, waiting for his opinion.

pbatard:
Still can't accept this translation:

image1

image2

Please create a new pull request once you have fixed these issue.

PR #1029, opened 2017-09-24
Request for Esata Hard drive Dock recognition

Geeknificent:
Hello I have been using yoursoftware quite successfuly for a few years now. But I have found that there is no option for Esata hard drive dock recognition. Can you make this a feature? I would love to use rufus to create images on hard drives in my Hard drive dock.


pbatard:
Not sure why you created this weird pull request.

I believe what you ask is already there, but as a __super-unofficial cheat mode__. Please try the <kbd>Ctrl</kbd>-<kbd>Alt</kbd>-<kbd>F</kbd> in Rufus. You may also have to enable Enumeration debug with <kbd>Alt</kbd>-<kbd>.</kbd>.

Note that this is only meant to work if your drive is seen as removable by the BIOS/UEFI firmware (so you may have to adjust your settings there, and if you can't then you're out of luck).

Now, since I have no plans to enable eSata outside of the current __UNSUPPORTED__ cheat mode (i.e. if it works, great, if it doesn't, you're on your own), I will now close this request.

PR #1033, opened 2017-10-03
[misc] add option to specify dd image

tgiphil:


pbatard:
I don't believe that's needed. It's only the UI that splits ISO and DD images as different beasts, but you can feed a DD image in the ISO image selection dialog, and Rufus will be smart enough to detect that it's a DD image and process it accordingly.

Have you  actually tried using `-i` with a DD Image?

Now, if you want to force an ISOHybrid image to be written in DD mode, that's a different story, but since `-i` should work for ISO/DD image selection, I don't think there is any need to add a new image selection option.

src/rufus.c
```diff
@@ -2966,6 +2977,8 @@ static void PrintUsage(char* appname)
 	printf("     Start in GUI mode (disable the 'rufus.com' commandline hogger)\n");
 	printf("  -i PATH, --iso=PATH\n");
 	printf("     Select the ISO image pointed by PATH to be used on startup\n");
```
JonnyTech: This should read `IMAGE FILE` instead of `ISO image` to cater for both ISO and DD image files (although the `--iso=` could be changed to `--image=` or just add it to keep backward compatibility)

pbatard: I'm planning to do just that in 3.0.

pbatard:
I'm going to close this pull request then.

PR #1034, opened 2017-10-03
Update rufus.loc Lithuanian translation

embar-:
https://github.com/pbatard/rufus/pull/1019#issuecomment-333821412 should be fixed now


pbatard:
Applied. Thanks.
Note that I prefer when patches are grouped, rather than split in 3 parts...

PR #1040, opened 2017-10-12
[cmp] propagate decompression errors from bled

wjt:
If, for example, you have a truncated gz-compressed file and try to write it to disk, `bled_uncompress_with_handles()` will return an error.  Previously, this was not reported back to the user -- there's no indication of any problem. If they're lucky, booting the partially-written disk will fail in an obvious way, but better to report an error sooner, even if we cannot give details.

(This is inspired by a similar change in [our fork](https://github.com/endlessm/rufus/blob/05d26bfe894b2058380c066ae08ae61950f51b0e/src/format.c#L1497-L1506). This branch is untested until Visual Studio 2017 finishes downloading!)

### DIFF
--- a/src/format.c
+++ b/src/format.c
@@ -1541,10 +1541,17 @@ static BOOL WriteDrive(HANDLE hPhysicalDrive, HANDLE hSourceImage)
 	LastRefresh = 0;
 
 	if (img_report.compression_type != BLED_COMPRESSION_NONE) {
+		int64_t bled_ret;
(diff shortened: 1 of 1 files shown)


wjt:
I have, however, tested the same change applied to our fork with our UI.

pbatard:
Very good point. Rufus should definitely report errors from Bled. The pull request has been applied &mdash; many thanks for this!

PR #1050, opened 2017-10-27
fix V773 from PVS-Studio

ip-gpu:
I'm a member of the Pinguem.ru competition on finding errors in open source projects. A bug, found using PVS-Studio. Warning: [V773](https://www.viva64.com/en/w/v773/) The function was exited without releasing the 'hDC' handle. A resource leak is possible.

(https://github.com/pbatard/rufus/blob/master/src/stdfn.c#L638)


pbatard:
Applied. Thanks!

PR #1067, opened 2017-12-06
Update rufus.loc PT-BR translations

trinaldi:
Thanks to Ricardo Pereira da Costa for the review.
Cleaning up the mess from the previous PR. 

Sorry, Pete, I'm in between a new city/job.


trinaldi:
Cleaning up the mess from the previous PR.

pbatard:
Thanks you very much for the quick refactoring of your pull request - much appreciated!

Everything looks good, so I have applied it.

The only small modification I applied was a change for _"Windows Para Levar"_ &rarr; _"Windows To Go (Windows Para Levar)"_ as my understanding is that Microsoft is using the _Windows To Go_ moniker in most regions, including the ones that don't speak English. Searching for _Windows Para Levar_ didn't seem to return many hits were it was used without also adding _Windows To Go_, so I prefer to have both terms used just in case. I'm hoping this small change can make things clearer about what this option does, and that this is okay with you.

Many thanks again for improving the translation.

trinaldi:
I agree, the old translation used _Windows To Go_, but since the reviewer use the translated version, I let it pass. But I definitely agree with you.

Thanks and sorry about some inconvenience .

PR #1113, opened 2018-04-19
[loc] Update Persian translation to latest

ziaa:


pbatard:
Applied. Much appreciated as always! 😄

Issue #296, opened 2014-03-11
Windows XP [Pro] (x64) Bootloader Detection


CJFWeatherhead:
Rufus currently doesn't detect the bootloader for Windows XP x64, for any of the Vanilla Technet Images (VL [with SP2], Retail/OEM [Gold] or the 2003 Preview). 

I have tested if Rufus detects Server 2003 [with or without R2] x64 [Editions] and it doesn't detect that either (XP x64/Server2k3 x64 share bootloaders so no surprises there.)

Using Rufus 1.4.5.428 on Windows 8.1 Pro [x64]

Please fix as I <3 Rufus :+1:

Issue #1095, opened 2018-03-08
[INTERNAL] List of 3.1 enhancements and fixes


pbatard:
__REMINDER: This topic is flagged as INTERNAL and is therefore NOT open for comments.__

General
----------
- [ ] Windows 10 type notifications for Rufus updates
- [ ] Remove XP ISO support?
- [ ] Full extraction of `efi.img` data (See `Solus-3-Budgie.iso`)

PR #1167, opened 2018-06-22
[iso] Support XP/Server 2003 x64 ISOs

Mattiwatti:
First of all, thanks for all the work you've put into Rufus. I've been using Rufus for years now and it's a been a breath of relief compared to the dark ages of things that came before it (I had many in-head acronyms for HPUSBFW, but the 'F' always stood for the same thing). So, apologies in advance for the fact that this PR probably made you groan as soon as you read the title, but... here it comes anyway.

Because of reasons, I needed to install a copy of XP x64 on an older machine yesterday. It also happened that the only DVD drive in my house (which I normally would have used, since XP's boot loader is not known for its USB-friendliness to begin with) was broken. So I fired up Rufus, and saw this for the first time in my life:

wtf

Even more amazingly, it turns out someone else had this same issue once (#296), although admittedly, that was over 4 years ago, and even then they were told to either fix it themselves or use a supported OS. The FAQ is also pretty clear w.r.t. XP support. So I'm mostly submitting this because I was going to write the patch regardless (I really have no choice in target OS for this machine) and if it gets merged and helps the other 3 people on this planet who use XP x64 for some reason - great. If not, that's fine too.

Some notes on this patch:
- The boot process of an x64 CD of XP/Server 2003 is very similar to Server 2003 x86 since they use the same 32 bit setup (by same I mean bitwise identical). Therefore most of the files relevant for Rufus are still under `\i386`, except for `txtsetup.sif` which is under `\amd64`. `setupldr.bin` must also be patched to use `\amd64` paths rather than `\i386` ones.
- The `winpe` field of `RUFUS_IMG_REPORT` has been widened to 2 bytes since there are now 3 base directories that may exist on a WinPE 1.x disk (`\i386`, `\amd64` and `\minint`) making the maximum possible flag value `>1` byte. This could probably be improved if needed since there are still really only `3^3` relevant files at most.
- `WINPE_MININT` is still mutually exclusive with `WINPE_I386` as well as the new `WINPE_AMD64`, but this is not true for `(WINPE_I386 & WINPE_AMD64)` which gives the nonzero `0x9`. Therefore these two flags should always be tested with a strict equality operator if the distinction is important.

### DIFF
--- a/src/format.c
+++ b/src/format.c
@@ -1167,19 +1167,22 @@ static BOOL WritePBR(HANDLE hLogicalVolume)
 static BOOL SetupWinPE(char drive_letter)
 {
 	char src[64], dst[32];
-	const char* basedir[] = { "i386", "minint" };
+	const char* basedir[] = { "i386", "amd64", "minint" };
 	const char* patch_str_org[] = { "\\minint\\txtsetup.sif", "\\minint\\system32\\" };
-	const char* patch_str_rep[] = { "\\i386\\txtsetup.sif", "\\i386\\system32\\" };
-	const char *win_nt_bt_org = "$win_nt$.~bt", *win_nt_bt_rep = "i386";
+	const char* patch_str_rep_i386[] = { "\\i386\\txtsetup.sif", "\\i386\\system32\\" };
+	const char* patch_str_rep_amd64[] = { "\\amd64\\txtsetup.sif", "\\amd64\\system32\\" };
+	const char **patch_str_rep;
+	const char *win_nt_bt_org = "$win_nt$.~bt", *win_nt_bt_rep_i386 = "i386", *win_nt_bt_rep_amd64 = "amd64", *win_nt_bt_rep;
 	const char *rdisk_zero = "rdisk(0)";
 	const LARGE_INTEGER liZero = { {0, 0} };
 	char setupsrcdev[64];
 	HANDLE handle = INVALID_HANDLE_VALUE;
-	DWORD i, j, size, rw_size, index = 0;
+	DWORD i, j, size, rw_size, index = 0, lst_index = 0;
 	BOOL r = FALSE;
 	char* buffer = NULL;
 
-	index = ((img_report.winpe&WINPE_I386) == WINPE_I386)?0:1;
+	index = (img_report.winpe&(WINPE_I386|WINPE_AMD64))?0:2;
+	lst_index = ((img_report.winpe&WINPE_I386) == WINPE_I386)?0:((img_report.winpe&WINPE_AMD64) == WINPE_AMD64)?1:2;
 	// Allow other values than harddisk 1, as per user choice for disk ID
 	static_sprintf(setupsrcdev, "SetupSourceDevice = \"\\device\\harddisk%d\\partition1\"",
 		ComboBox_GetCurSel(hDiskID));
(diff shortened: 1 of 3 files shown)


pbatard:
Agh!! Where were you 3-4 years ago, when people were pestering me about once a month with regards to XP64 or 2003 installation support?

This patch looks straightforward enough, and considering that I have not yet removed the ability to create XP bootable USBs (you're quite lucky about that, because that was almost the first thing I was about to do for the 3.1 release, as per #1095), I actually have no qualms about applying it, once I have tested it a little bit, since it's not going to affect a part of the code I care much about at this stage...

So you were right about trying your luck with a pull request, as I'm planning to integrate it to the code. I just wish you had submitted that one a few years earlier... 😄

Mattiwatti:
Where I was 4 years ago? Not sure, but clearly I was doing a better job of not having to install XP at the time than I am now :smile: I'm surprised you got so many requests for it to be honest - I guess there are a lot of masochists out there...

Seriously though, purely as an OS XP/2003 was actually very impressive at the time especially considering it was MS's first x64 OS. The company I worked for had a few 2003 servers running in production and they were rock solid. But all those machines contained only carefully selected (and expensive...) hardware that was known to actually have working x64 drivers available. For home use it was an outright disaster, and it still is. By the time vendors finally caught up with x64 support, XP was already out of date and many never even bothered releasing drivers for it.

I did see your comment on #1095 and it was basically my line of thinking as well. Is this patch useless? Kinda. Is it for an OS that's not officially even supported? Yeah. But in terms of damage done it only comes in at +6 LOC at the moment which I'd say is acceptable. If I'd had to rewrite, say, all of `SetupWinPE()` from scratch only to add x64 support I wouldn't have submitted a PR for that.

blackcrack:
well, Reactos it is at moment on the way to be a modern XP/S2003 (and it is for free/for no money/for less as nothing ... and, the best , reactos is the sourcecode free/open and .. you can helping with !!
And.. the companys can helping with to create a free/open Source and honest Orderingsystem from the Community for the community of the internet.. and the way it doesn't stop behind version 1.0, goes further up to Win7 and later (what the community determine) and -not- a M$ Designer 
The compatibility from (if anyone implement 3.11) from WinXP/S2003 up to .... should stay ad infinitum..
or so.. and should support for ever skining like it was in XP and because the open sourcecode it is possible for programming patches to change it like the owner it want  .. and can do all what the user/owner want..
so i see no different between Reactos and WinXP.. the game goes far, further,  far away..... outer spaceeeee ;) (joke) but it becomes really interesting, just because it is possible to building jobs in Factorys and the Factorys can have his own programming employee('s) and can have more Data security and Factory Security because, the information's can stay in the factory's and in the land's of the factory's now..
so, it is very important to supporting Reactos.. it make jobst and helps to build up :)

best regards
Blacky

pbatard:
Now applied.

Note that I modified your commit a little bit to use multidimensional tables (cleaner than switching table pointers IMO), and I also reworked the `WINPE_###` bitmap constants so that the bits are grouped by "arch" (`i386` &rarr; 3 least significant bits, etc.), so that the bit mapping is less confusing.

I only ran some __limited__ testing with Win2k3 x64 (no XP x64), as well as with a regular XP and MiniNT ISO to confirm that there was no regression.

If you feel like it, you can use https://rufus.ie/testing/rufus-3.1_TEST1.exe that includes the modified commit, as well as your other commit with the speed improvements.

Thanks again for the patch!

Mattiwatti:
Cool. Your method is better. I actually wanted to do the exact same thing (using multidimensional arrays) but decided against it in order to keep the existing code mostly undisturbed. But this way is more readable and also provides future proofing for when the 128 bit version of Windows XP is released.

>I only ran some limited testing with Win2k3 x64 (no XP x64)

Windows 2003 and XP x64 are **exactly** the same thing, so either would have worked :) To be precise, XP x64 RTM corresponds to Server 2003 SP1 (the RTM version of Server 2003 was only available for x86 and Itanium). Of course Microsoft has always used shared codebases for server and client editions (this is only logical), but it's not unfair to say that XP x64 really is just Server 2003 with a different boot logo, themes enabled by default and of course the amazing [64 bit version of Pinball](https://blogs.msdn.microsoft.com/oldnewthing/20121218-00/?p=5803).

I tried your test exe and both formatting and booting the stick worked without any issues (besides the ones inherent to XP x64 in general I mean...). So all seems OK.

pbatard:
Great! Thanks for testing.

PR #1174, opened 2018-06-28
add cheat mode (alt-"-") to quickly erase a flash drive 

renevdzee:
I add a cheat mode (alt-"-") to quickly erase a flash drive by skipping blocks that are already empty (a usb flashdrive reads a lot faster than it writes). This is very useful to erase partially used flash drives.

* classifies blocks filled with either 0x00 or 0xff as empty. Some/most flash drives are erased to 0xff from the factory.
* added a new message MSG_306, so far only translated into dutch.
* added a BOOL "smart_skip" to enable this feature. Currently setting "zero_drive" combined with "smart_skip" enables the erase cheat mode. I am not sure if this is the right way. 
* could also be enabled to skip unchanged blocks when writing or zeroing a drive, but I don't see the use-case for that.


JesseNaser:
# l "es-ES" "Spanish (Español)" 0x040a

t MSG_306 "Borrando disco: %0.1f%% completado"

pbatard:
Okay, I have now applied a slightly modified version of this patch to Rufus mainline at long last.

Note that I have changed the cheat mode from, <kbd>Alt</kbd>-<kbd>-</kbd> to <kbd>Ctrl</kbd>-<kbd>Alt</kbd>-<kbd>Z</kbd>, since this is pretty much an enhancement of <kbd>Alt</kbd>-<kbd>Z</kbd>, and I am already doing something like this for <kbd>Alt</kbd>-<kbd>F</kbd>/<kbd>Ctrl</kbd>-<kbd>Alt</kbd>-<kbd>F</kbd>.

I am also renaming this feature to _Fast-zeroing_, even as I understand that there are provisions to use it for more than zeroing a drive. The reason behind that is that I really want users who may only get a clue about the feature from what is displayed in the UI to at least get some idea of the __intent__ behind this feature, which I don't thing "Erase" vs "Zero" provides much insight about (even if, depending on your hardware and existing content, _Fast-zeroing_ may actually be slower than regular zeroing).

As a result of this, I had to ditch the NL and ES translation. But to tell the truth, I very much __prefer__ to have the regular translator take care of translations, for many different reasons, one of which being that, until I start a new round of translation (which is something I try to limit at no more than one every 5-6 months, as it gets exceedingly time consuming for me otherwise), I want to be able to change any new/modified English messages freely, which means that none of the new English changes should be considered final. So please don't worry about providing me a translation for a single message. I would very much prefer that you do not.

At any rate, thank you very much for this pull request!

PR #1182, opened 2018-07-21
[loc] update Traditional Chinese translation to latest

Chocobo1:
A bit of problem here, pardon... Since my poedit is installed on a linux machine,  I don't know how tell `pollock.exe` to convert zh-TW.po back to rufus.loc file, so I'm only sending you the .po file here.
Let me know if there is any way I can do better.

### DIFF
--- a/res/loc/po/zh-TW.po
+++ b/res/loc/po/zh-TW.po
@@ -1,0 +1,1596 @@
+msgid ""
+msgstr ""
+"Project-Id-Version: 3.2\n"
+"Report-Msgid-Bugs-To: pete@akeo.ie\n"
+"POT-Creation-Date: 2018-07-21 20:46+0800\n"
+"PO-Revision-Date: 2018-07-21 22:21+0800\n"
+"Language: zh_TW\n"
+"MIME-Version: 1.0\n"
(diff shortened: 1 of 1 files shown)


Chocobo1:
>I can look into creating a .NET core version that can work there, but I'd like to understand your requirements a bit better.

Running on linux is an edge case, so I won't bother if I were you.

However it might improve the situation if the conversion back to .loc is done when I see the message "Poedit was not found. You will have to launch it and ..." and "Now press any key to launch your e-mail client ...". I mean do the convert after key input (just before the e-mail client is launched).
I'm not sure it will make sense, just some idea.

pbatard:
Actually, what you want should already be in there. The process that monitors whether the `.po` has been updated is launched regardless of whether Poedit is found or not.

So, if you copy the `.po` back to the location where you picked it, and you still have Pollock running, it should create the updated `.loc` for you, even if you didn't run Poedit on the same machine.

I guess you can try that with the next round of translations, and let me know if that doesn't work.

Chocobo1:
>Actually, what you want should already be in there. 

Tested, I can verify it do work as expected! :+1:

PR #1184, opened 2018-07-22
[loc] update Portuguese Brazilian translation to latest

trinaldi:
Translations are now easier thanks to the new system. This is the new version, with `.po` file created.

Keep it up.


pbatard:
Looks good. Thanks!

PR #1236, opened 2018-10-23
Add command line handling of preferred file system.

chazste:
Pete,
     Here is my proposed implementation of a command line parameter to set up a preferred filesystem type. I modeled the command line handling off of the existing -l code. I believe it works the way you requested. The only thing I noticed is that even though this was intended (by me anyway) to be used with -i it stays in effect for the entire "session" while -i is a one-shot.  I think that is OK and the both options have the behavior that I would expect. But that may just be because I don't want to write the one extra line to make this a one-shot as well :)

    Also I did not choose to mess with your existing one character command line options. I just chose -p for "preferred file system" and called it good. We had discussed -d for default but I think preferred is a better term anyway based on the behavior.

### DIFF
--- a/src/rufus.c
+++ b/src/rufus.c
@@ -114,6 +114,7 @@ char embedded_sl_version_str[2][12] = { "?.??", "?.??" };
 char embedded_sl_version_ext[2][32];
 char ClusterSizeLabel[MAX_CLUSTER_SIZES][64];
 char msgbox[1024], msgbox_title[32], *ini_file = NULL, *image_path = NULL, *short_image_path;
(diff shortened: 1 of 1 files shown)


pbatard:
Thanks. It may be a few days before I get a chance to look into it (and I am still planning to switch the option to `-f`, as I'm pretty sure this will be a lot more intuitive for an option that effectively sets the file system than relying on people remembering that "preferred" qualifier...

pbatard:
Applied, with a few changes. Mostly, I set the new option to use the `-f` flag, as I mentioned, and moved the old `-f` to `-x`.

Many thanks.

PR #1244, opened 2018-11-10
issue 1243

chazste:
Move console attach closer to the creation of the hogger, and only even attempt the attach if the user did not request a GUI program.

Peter,
     This change is a little more disruptive than I had originally envisioned. However I think this is a good compromise between the size of the change and keeping the intent clear. The two operations that deal with the parent process' console are now both protected by the -g flag.


pbatard:
Looks good to me. Thanks for the patch.

PR #1293, opened 2019-03-29
Updating Turkish localization

Tmp341:
- Fixing typos
- Adding untranslated parts
- Making small changes, so they would be more understandable.

You can wait for @MehmetaliKuran 's approval.


pbatard:
Thanks to you both. I have applied the update.

PR #1294, opened 2019-03-31
Fix grammar

marcellogianola:


marcellogianola:
sorry, I wrote grammar but I mean orthography...

pbatard:
Applied - Thanks!

PR #1334, opened 2019-06-22
Added Albanian language translation

xray700r:
Added Albanian language translation in .po file format. Full translation with official Albanian language. No specific dialects used.

### DIFF
--- a/res/loc/po/al-AL.po
+++ b/res/loc/po/al-AL.po
@@ -1,0 +1,1731 @@
+msgid ""
+msgstr ""
+"Project-Id-Version: Albanian translation 1.0\n"
+"Report-Msgid-Bugs-To: pete@akeo.ie\n"
+"POT-Creation-Date: 2019-03-19 08:16+0100\n"
+"PO-Revision-Date: 2019-06-22 22:30+0200\n"
+"Language: sq_AL\n"
+"MIME-Version: 1.0\n"
+"Content-Type: text/plain; charset=UTF-8\n"
+"Content-Transfer-Encoding: 8bit\n"
+"X-Poedit-SourceCharset: UTF-8\n"
+"X-Rufus-LanguageName: Italian (Italiano)\n"
+"X-Rufus-LCID: 0x0410, 0x0810\n"
+"Last-Translator: \n"
+"Language-Team: Moisi Xhaferaj <elektroxray@gmail.com>\n"
+"X-Generator: Poedit 1.8.9\n"
+
+#. • IDD_DIALOG → IDS_DRIVE_PROPERTIES_TXT
+msgid "Drive Properties"
+msgstr "Vetite e Drive"
+
+#. • IDD_DIALOG → IDS_DEVICE_TXT
+msgid "Device"
+msgstr "Aparati/Njesia"
+
+#. • IDD_DIALOG → IDS_BOOT_SELECTION_TXT
+msgid "Boot selection"
+msgstr "Zgjidhja e Boot"
+
+#. • IDD_DIALOG → IDC_SELECT
+msgid "Select"
+msgstr "Zgjidh"
+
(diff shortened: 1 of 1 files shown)


pbatard:
I hate to do that, but I'm afraid that I am going to have to reject that request.

As explained [in the FAQ](https://github.com/pbatard/rufus/wiki/FAQ#What_languages_are_natively_supported_by_Rufus), the only languages that I accept for localization of the Rufus applications are languages for which a fully localized version of Windows exists. And if you look at the _Language versions_ for full locations of Windows [here](https://www.microsoft.com/en-us/windows/windows-10-specifications#primaryR6), you'll see that, unfortunately, Albanian is not on that list.

Also, translating Rufus to a new language is not a matter of just adding a new ` .po`. You are supposed to follow [this guide](https://github.com/pbatard/rufus/wiki/Localization) and __fully validate__ your translation through the use of Pollock (which will convert the ` .po` to a local `.loc`  you can test). And you should always contact the developer before you perform the work.

Now, you may be wondering: _"But what harm could adding a new language do, when someone's already done all the work?"_, to which I will indicate that:
* Maintaining translations actually has a large hidden cost for me, as, every time there is a new translation update round (the application is far from being static, and every time I add or modify features, I tend to need a new bunch of translation updates), I end up spending __A LOT__ of time chasing after translators and validating their updates. I quickly found out that having to do that for more than 30 languages takes a lot more time away from development than I anticipated, so I definitely want to limit the number of languages Rufus is translated into.
* I did that for Azerbaijani, and this is currently creating an issue, as the original translator is now M.I.A. and despite repeated requests for help with a new translator (which you can see on the https://rufus.ie website), I'm finding that nobody is willing to take that job over, which is problematic, as the Azerbaijani translation is now trailing so far behind that I don't think it's worth proposing to Rufus users and unless someone does volunteer to help __soon__, I will most likely drop Azerbaijani altogether from the 3.6 release.
* Adding one language simply because _"someone did the work"_ would open the floodgates to people proposing localization for their own language or jockeying for preferential treatment (_"Why did you accept to add Albanian but not my language?"_), which I'd rather avoid especially when I really want to limit the number of languages I have to maintain for the application.

Therefore, despite your kind proposal and, no doubt, all the hard work you put in it, I see no other alternative but to refuse this pull request. I'm very sorry to have to do that, but I hope that you can understand the reasons I have for doing so. I will also advise you, next time you want to propose to help a project, to maybe try to contact its developers first, to see if they will be able to accept your proposal, as one need to be careful to see Open Source projects as places that will accept new feature after new feature. Instead, I would assert that, a well maintained Open Source project will often set limits with regards to what they are ready to accept as contributions, and what they will reject, so that the project remains manageable, and therefore, if you are planning to invest a good chunk of your time on a proposal, you may want to check that it doesn't fall into the "will be rejected" category.

So, once again, I'm really sorry to have to reject this pull request, and I especially hope that it won't discourage you from wanting to contribute to other Open Source projects, as we can always use some help!

xray700r:
Thank you for your clarification. I indeed read the FAQ and the rest of the documentation but did not notice only the only important note in which you define the fully supported localized languages. That is an attention issue I would say since I didn't notice it and is my problem. I'm not requesting anything particular i did my work and I liked it once I have some free time I will compile Rufus and try it on Windows 10 to see if it works in an not fully localized language. 
When I commit to Open Source projects it is not because I want to, most of the time it is to give back something when you've used a great software tool so many times. I did my part, even if it is not useful to you. Good luck with the other 30+ languages!!!

pbatard:
Thanks for understanding. I can only commend you for being willing to give something back the Open Source projects you find useful &mdash; We definitely need more people like you!

I hope this experience won't put you off from using Rufus. I promise I'll do my best so that you continue to find Rufus useful in the future (even if its UI is not in Albanian)...

PR #1341, opened 2019-07-12
Update it-IT.po

marcellogianola:


pbatard:
Rejected. Please don't edit localization files by hand, as you will introduce mistakes (which you did here). Instead you should follow the procedure highlighted [here](https://github.com/pbatard/rufus/wiki/Localization#Editing_an_existing_translation) and, especially, __validate__ your changes in Rufus, as, for instance, some messages may have to be kept under a certain length for the UI to look good.

PR #1355, opened 2019-08-02
ASN.1 parser's sequence tag is wrong

gusrb406:
- Sequence is encoded with tag byte of 0x30.
- https://docs.microsoft.com/en-us/windows/win32/seccertenroll/about-sequence

### DIFF
--- a/src/parser.c
+++ b/src/parser.c
@@ -1276,7 +1276,7 @@ static BOOL get_data_from_asn1_internal(const uint8_t* buf, size_t buf_len, cons
 	BOOL is_sequence, is_universal_tag;
 
 	while (pos < buf_len) {
-		is_sequence = buf[pos] & 0x20;
+		is_sequence = buf[pos] & 0x30;
 		is_universal_tag = ((buf[pos] & 0xC0) == 0x00);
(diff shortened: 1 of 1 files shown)


pbatard:
Are you sure about this?

According to https://en.wikipedia.org/wiki/X.690#Encoding, bits 8-7 (mask `0xC0`) are the tag class, bit 6 (mask `0x20`, i.e. the one we are interested in here) is the toggle between Primitive and Constructed (in other words, non-sequence or sequence) and the other bits (mask `0x1F`) are the tag number.

So I believe checking if bit 6 is set, by masking against `0x20`, is the proper way to determine if we have a sequence or not.

On the other hand, using `0x30` for sequences means that we expect all sequences to have a tag number of 16 (`0x10`) or higher, and I don't see how you would patch the sequence mask to `0x30` if you don't also patch the tag mask 2 lines further down to `0x0F` to be consistent.

Can you describe the actual problem you're observing that prompted you to submit this PR? How exactly does using masking against `0x20` create an issue in what you are trying to accomplish?

pbatard:
Since I haven't heard back on this pull request, I'm going to close it.

PR #1424, opened 2019-12-21
leave .efi files in efi/boot/ uncompressed on ntfs

josephsmendoza:
additional fox for #1381 on top of commit 8719412667776c0be98c4036ee3fb562d008e7bc


pbatard:
Applied, thanks.

PR #1426, opened 2019-12-23
Update de-DE.po

Mr-Update:
German translation improved


pbatard:
Sorry for the late reply. Your update has now been applied &mdash; Many thanks for it!

pbatard:
@Mr-Update.

Please be mindful that you introduced regressions!

For instance you changed:
```
t MSG_217 "Formatieren: %s"
```
back to
```
"Formatieren: %0.1f%% erledig
```

This actually results in German users complaining that the percent display no longer works in Rufus 3.9!

Next time you edit a translation, do not use the `.po` files from the repository but use `pollock.exe` as documented [here](https://github.com/pbatard/rufus/wiki/Localization#Editing_an_existing_translation). Thank you.

PR #1438, opened 2020-01-26
Add User Specified default for --gui command line option

chazste:
@pbatard 
I would like to be able to persist my desire to run with the --gui command line option enabled most of the time. Without having to run a custom version of rufus (built from source). This change should be minimally disruptive (it modifies 2 lines and adds one). It maintains the current behavior in all "normal" situations. 

## Considerations
I originally intended to use the rufus specific ReadSettingBool() function, but I didn't for two reasons: one is that this is read so early that the ini file has not been initialized yet so it would always read from the registry; two is that if the key does not exist in the registry it would be written even when running the portable version of rufus. This leaves me using a SHLWAPI function "bare". The good news is that it is read only. The bad news is that it will also look from the key under HKLM as well as HKCU.

Also I originally was hoping to use --no-gui and -g- to turn this feature back off. Unfortunately the C version of getopt_long does not support this feature (other language implementations like ruby and node seem to) also this does not work since -g is pre-processed anyway. Therefore I chose to make -g be a toggle. This has minimal impact on code size but may be seen as counter-intuitive.

## Code size
My test of the x86 release build saw no change in size. However it does not appear that the .sln file applies UPX, and the default code alignment is 4K I think; so not too surprising. 

If you are willing to except this ER but wish to have these concerns addressed in a different way I will be happy to try again.

### DIFF
--- a/src/rufus.c
+++ b/src/rufus.c
@@ -2873,7 +2873,8 @@ int WINAPI WinMain(HINSTANCE hInstance, HINSTANCE hPrevInstance, LPSTR lpCmdLine
 	int wait_for_mutex = 0;
 	FILE* fd;
 	BOOL attached_console = FALSE, external_loc_file = FALSE, lgp_set = FALSE, automount = TRUE;
-	BOOL disable_hogger = FALSE, previous_enable_HDDs = FALSE, vc = IsRegistryNode(REGKEY_HKCU, vs_reg);
+	BOOL disable_hogger = SHRegGetBoolUSValueA("SOFTWARE\\" COMPANY_NAME "\\" APPLICATION_NAME, "gui", /*fIgnoreHKCU*/ FALSE, /*fDefault*/ FALSE);
+	BOOL previous_enable_HDDs = FALSE, vc = IsRegistryNode(REGKEY_HKCU, vs_reg);
 	BOOL alt_pressed = FALSE, alt_command = FALSE;
 	BYTE *loc_data;
 	DWORD loc_size, u, size = sizeof(u);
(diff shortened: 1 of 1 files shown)


pbatard:
Can you please describe the nature of problem you are trying to solve with this?

Right now I fail to see why you need a setting to replace `--gui` when that option is meant to be used for command line usage only, and if you do have command line in the first place, then it's hard to see why you need to have something else to supersede `--gui`.

Please bear in mind that I am __exceedingly__ reluctant to add features that are only going to help a very limited number of users, even if they boil down to a couple of lines in the code, because designing Rufus for super-specific scenarios that only a handful of people are going to encounter is the shortest path to maintenance hell...

I could certainly see `--gui` applying to various console users, which is why I had not much of an issue adding that option, but, short of having a case for it that would explain how what you propose is going to benefit a large enough number of users, this change does appear to me like something that's designed specifically for your unique usage, and therefore can't see myself applying it without a solid justification of how it will help others (because it I were to open the door here, then I'd get submerged with requests of _"But you added a feature for usage that's unique to @chazste, so why can't you add one for mine too!!!"_).

So, what's your use-case for this feature? And what specific is it solving exactly?

TrainsDontHunt:
Not a fan of this logic:

`disable_hogger ^= TRUE;

It assumes you know the values of TRUE and FALSE, which you can't rely on. Either use zero and one so you know, or something like this:

`disable_hogger = !disable_hogger;

pbatard:
@chazste, can you please answer the question I asked? If not I will close this PR.

chazste:
@pbatard, sorry for the delay. It's been a busy week :(

After thinking about this some more; I am going to close this PR. In addition to your concern about the number of users that would benefit from this option, I think the fact that it does not integrate nicely with the existing option processing is also a big reason not to do this.

Thanks for your consideration.

PR #1449, opened 2020-02-09
Bump the autotools versions and add support for an out-of-tree build.

Davester47:
Most open source projects support an out-of-tree build, where the src directory and the build directory aren't the same, but I had problems getting it to work with rufus. I ran into some missing header files, that I found in the source tree, but they weren't being included. I looked in the Makefile.am's and found that they were using CFLAGS like: ```-I. -I..``` instead of ```-I$(srcdir) -I$(srcdir)/..``` I modified those as well as the Makefile.am in res/loc for a similar purpose. I also found that I coudn't compile rufus without specific older versions of automake and aclocal, so I ran autoreconf. I wouldn't have included the updated Makefile.in's in this PR, but they were affected by my Makefile.am changes.

By the way, good job on Rufus! I wish I had found it earlier.


Davester47:
I just realized that there's a problem in my PR. The application can't start, since it can't find locales, and I'm working on fixing it.

Davester47:
You probably aren't going to accept this PR now, but my issue was that windres couldn't find the included rc files without knowing where the source directory was at compile time. I made rufus.rc an output file of configure, which fills in the blanks inside of rufus.rc, and it now builds with no issues. I don't know if this breaks compatibility with Visual Studio, though, since it looked like that already generated it.

pbatard:
Yeah, this looks like a lot of alterations, some of them breaking, just so a couple people can build out-of-tree, for benefits that are very hard to grasp (coz, with Rufus being a standalone app rather than something you'd use in other projects, I really fail to see why out-of-tree builds are that important to have).

I appreciate the work you did, and I may still pick the `.am`/`.ac` changes you applied, but I can't go with your `.rc` changes, since they completely break MSVC compilation.

As such, I will close this PR.

PR #1454, opened 2020-02-13
[core] Only retry writes when they fail

ManMower:
ClearMBRGPT attempts to write WRITE_RETRIES times, even if all those
times succeed.

Instead, skip the remaining retries on success.


pbatard:
Well spotted! I certainly screwed up these loops when I wrote them, so thanks for submitting a PR to fix that.

I've applied your patch, after modifying it to make the exit condition clearer as well as harmonizing the code a little bit. Thanks again.

PR #1459, opened 2020-02-18
Fix some PVS-Studio static analyzer issues

dimhotepus:
Issues (some, stay tuned) found by PVS-Studio static analyzer.

Product page: https://www.viva64.com/en/pvs-studio/
How to use for free: https://www.viva64.com/en/b/0600/

### DIFF
--- a/src/ext2fs/nt_io.c
+++ b/src/ext2fs/nt_io.c
@@ -283,7 +283,7 @@ static __inline BOOLEAN _IsMounted(IN HANDLE Handle)
 	NTSTATUS Status = STATUS_DLL_NOT_FOUND;
 	PF_INIT(NtFsControlFile, NtDll);
 	if (pfNtFsControlFile != NULL)
-		pfNtFsControlFile(Handle, 0, 0, 0, &IoStatusBlock, FSCTL_IS_VOLUME_MOUNTED, 0, 0, 0, 0);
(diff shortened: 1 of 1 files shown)


pbatard:
1) Yes
2) Not unless you send a similar patch to [the official e2fsprogs](https://github.com/tytso/e2fsprogs/blob/master/lib/ext2fs/fileio.c#L359-L366) and they integrate it. I really want to avoid deviating from the e2fsprogs upstream code unless I have a good reasons to do so, and keeping PVS happy is not one of them. I also expect that e2fsprogs is doing it this way for a reason...
3) I'm afraid you broke MinGW builds, which means I can't accept this patch. Moreover, if you want to submit patches against Bled, I would prefer if you did that against [the upstream source](https://github.com/pbatard/bled) as I'd rather see patches go downstream (Bled &rarr; Rufus) than upstream. But that is not that big a deal, as long as you fix MinGW builds.

dimhotepus:
Decided to send last two patches to related upstreams. Thanks for suggestion!

As for first one, it is already fixed in [e2fsprogs upstream](https://github.com/tytso/e2fsprogs/blob/master/lib/ext2fs/nt_io.c#L580-L582), so feel free either to apply it, or just upstream `e2fsprogs` dependency.

pbatard:
Applied. Thanks!

PR #1464, opened 2020-02-23
Fix PVS-Studio static analyzer issues #2

dimhotepus:
Issues found by PVS-Studio static analyzer.

Product page: https://www.viva64.com/en/pvs-studio/
How to use for free: https://www.viva64.com/en/b/0600/

### DIFF
--- a/src/badblocks.c
+++ b/src/badblocks.c
@@ -463,7 +463,7 @@ static unsigned int test_rw(HANDLE hDrive, blk64_t last_block, size_t block_size
 			if (max_bb && bb_count >= max_bb) {
 				if (s_flag || v_flag) {
 					uprintf(abort_msg);
-					fprintf(log_fd, abort_msg);
+					fprintf(log_fd, "%s", abort_msg);
 					fflush(log_fd);
 				}
 				cancel_ops = -1;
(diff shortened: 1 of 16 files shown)


pbatard:
I don't mind the warnings that are triggered by actual potential vulnerabilities or unexpected outcome. Some of these are very good finds indeed and I most certainly want to fix that.

But the micro-optimization ones line `strlen(x) == 0` vs `x[0] == 0` or computing token length only once are a bit too much. For instance, the duplicated token length was very deliberate for code readability, especially as one can expect the compiler to perform that kind of optimization behind the scenes, and it's not like we're in a loop that computes a SHA. So the "performance" gain here is completely pointless.

Likewise the duplicated `(line[i] == '"')` was intentional for clarity, since it matches the condition on the next line and the compiler will have the result of the previous check in a status register anyway, so whether you have `if ((a == b) || ((a !=b) && (c == d)))` or just `if ((a == b) || (c == d))`, it doesn't make much of an actual difference anyway, and I'm worried about optimization that takes away some of the code readability.

Still, this PR is great, so thanks a lot for submitting it! I will apply it when I get a chance.

dimhotepus:
Regarding performance vs readability, looks like you are right, performance gains may be negligible here.

Thank you for the great tool :)

pbatard:
I now have applied the changes that I saw relevant. I will therefore close this PR with the pushed commit.

PR #1550, opened 2020-05-30
Add http2, gzip, deflate support

dmex:
This PR enables support for the http/2 protocol and enables automatic gzip/deflate decoding support. 

These options are backwards compatible and are only used when the server supports these features.

### DIFF
--- a/src/missing.h
+++ b/src/missing.h
@@ -109,6 +109,15 @@ static __inline void *_reallocf(void *ptr, size_t size) {
 #ifndef WM_COPYGLOBALDATA
 #define WM_COPYGLOBALDATA 0x49
 #endif
+#ifndef INTERNET_OPTION_ENABLE_HTTP_PROTOCOL
+#define INTERNET_OPTION_ENABLE_HTTP_PROTOCOL 148
+#endif
+#ifndef INTERNET_OPTION_HTTP_DECODING
+#define INTERNET_OPTION_HTTP_DECODING 65
+#endif
+#ifndef HTTP_PROTOCOL_FLAG_HTTP2
+#define HTTP_PROTOCOL_FLAG_HTTP2 0x2
+#endif
 
 /* The following is used for native ISO mounting in Windows 8 or later */
 #define VIRTUAL_STORAGE_TYPE_VENDOR_MICROSOFT \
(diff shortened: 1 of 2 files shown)


pbatard:
Sounds fair enough, though this request currently breaks MinGW builds as it missing a few constants that are only defined in MSVS (see [here](https://ci.appveyor.com/project/pbatard/rufus/builds/33216880/job/t0eghne9s1voyche)). Also, you have a typo for `protocalSupport` &rarr; `protocolSupport` and to be consistent with the rest of the code, the variable should be `dwProtocolSupport`.

I would also define a global string constant for `"Accept-Encoding: gzip, deflate"` so that, if we ever add more than gzip, deflate, we only have one instance to modify.

dmex:
> breaks MinGW builds as it missing a few constants that are only defined in MSVC

I've added the types to the missing.h header.

> Also, you have a typo

Ah damn, fixed 😄 

> I would also define a global string constant for "Accept-Encoding: gzip, deflate" so that, if we ever add more than gzip, deflate, we only have one instance to modify.

This has been fixed. I also wanted to request enabling the "Enable string pooling" option for the project solution (C++> Code generation) ?

pbatard:
Looks good. Thanks a bunch for updating the PR. I'll apply it when I get a chance, but I'll probably do that after a staged commit that I currently have waiting.

> I also wanted to request enabling the "Enable string pooling" option

Huh, VS is not doing that by default? I definitely want it, so I'll do that, though the binaries I release are MinGW ones, and I'm assuming that the current gcc optimization is taking care of string dupes (though, seeing that VS2019 doesn't, I'm going to double check this as well).

Many thanks for the PR! 😁

dmex:
> VS is not doing that by default?

The default is automatic which doesn't enable depending on various options. Another option added recently is the Advanced > "Use debug libraries" which needs to be Yes for the debug configuration or the binaries don't link with the msvc debug libraries.

> current gcc optimization is taking care of string dupe

GCC does this by default last I checked 👍

pbatard:
Okay, I've enabled string pooling for Debug VS2009 builds and I've also made sure that `Use debug libraries` was set where needed (I only had it on some of the libraries).

I will close this as I push the your commit. Thanks again for this PR!

PR #1572, opened 2020-07-03
test

dyeske:
test ok this failed I did it backwards, please disregard


dyeske:
I messed this up, please disregard

pbatard:
I'd rather you justify why exactly you think you can't use an up to date MinGW.

`PROCESSOR_ARCHITECTURE_ARM64` is defined in up to date MinGW, so if you have an issue with it not being defined, please contact the maintainers of your distro to ensure they update their MinGW package to have it.

Else, you are going to create pull request after pull request for all the MinGW projects you want to compile for ARM64, which is not the proper way to go.

The only way I can ever accept this is if you show proof that you have tried to get MinGW packages updated and that the maintainers of whatever distro you are using have provided a good reason not to do so.

Also, please understand that I regularly remove code that existed only for compatibility with older toolchain versions (case in point [from today's commit](https://github.com/pbatard/rufus/commit/500172a7a3604abd194cec984ce642c53cc280a9#diff-b1e6758f1628c57f86c1378d0c5e52a8L1357-L1360)), as I genuinely don't want to be bogged down with trying to support obsolete stuff. As such, even if a fix to support an old(ish) toolchain is a one liner, I will need good justification to add it, as I am literally fighting to remove that stuff from Rufus.

PR #1579, opened 2020-07-14
Delete index.html

Stylix58:
This folder contain the same index.html file of root folder. So, i am destroyed this folder because this is useless

### DIFF
--- a/downloads/index.html
+++ b/downloads/index.html
@@ -1,8 +1,0 @@
-<!DOCTYPE html>
-<html>
-<head>
-<meta http-equiv="refresh" content="0; url=https://rufus.ie/downloads">
-</head>
-<body>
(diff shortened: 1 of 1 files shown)


pbatard:
This is needed for the http to https redirection of the downloads, because we don't want to enforce blanket https always for various compatibility reasons.

For instance we want http://rufus.akeo.ie/downloads to redirect to https://rufus.ie/downloads while leaving other sections of the site accessible through http.

Please refrain from suggesting GitHub pages cleanup, so assuming that a file is "useless" unless you are very familiar with the history of this site as well as GitHub's hosting limitations.

Unless you are actually seeing an issue linked to using a specific file, you don't want to start removing them at random.

&rarr; Rejected.

PR #1580, opened 2020-07-14
Delete index.html

Stylix58:
This folder contain the same index.html file of root folder. So, i am destroyed this folder because this is useless


pbatard:
Same story as #1579

PR #1608, opened 2020-08-29
[ui] lossless optimization of icons

tiagofilipesilva:
- the icons were optimized in order to reduce the binary size without quality loss


pbatard:
I'll apply this when I get a chance (I've got a few things in the way of a clean-enough repo allowing me to proceed, so this may actually take a couple of weeks).

pbatard:
Applied. Thanks again!

PR #1633, opened 2020-10-12
docs: fix simple typo, tolerence -> tolerance

timgates42:
There is a small typo in src/libcdio/cdio/iso9660.h, src/libcdio/iso9660/iso9660_fs.c.

Should read `tolerance` rather than `tolerence`.

### DIFF
--- a/src/libcdio/cdio/iso9660.h
+++ b/src/libcdio/cdio/iso9660.h
@@ -638,7 +638,7 @@ typedef struct _iso9660_s iso9660_t;
     contained in a file format that libiso9660 doesn't know natively
     (or knows imperfectly.)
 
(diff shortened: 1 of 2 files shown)


pbatard:
That's a [libcdio](https://www.gnu.org/software/libcdio/) patch, not a Rufus patch (Please be mindful that Rufus might be reusing components and therefore, since we need to have the ability to fetch newer versions without applying extra changes on top of it, we cannot accept changes for elements that should be patched upstream).

Please send that patch to [libcdio-devel](https://lists.gnu.org/mailman/listinfo/libcdio-devel). Once it's been applied, it will naturally find its way into Rufus.

PR #1635, opened 2020-10-15
Update zh-CN.po

dingpengyu:
The specific updates are as follows:
将megabytes翻译维m 字节
修复 请稍等...
修复 运行下载脚本...
修复 启用涵盖 USB 硬盘的显示
请谨慎使用，后果自负!!!
翻译 KB条目
翻译 MB条目
翻译 GB条目
翻译 TB条目
翻译 PB条目
翻译 Windows To Go条目

### DIFF
--- a/res/loc/po/zh-CN.po
+++ b/res/loc/po/zh-CN.po
@@ -1,9 +1,9 @@
 msgid ""
 msgstr ""
-"Project-Id-Version: 3.5\n"
+"Project-Id-Version: rufus\n"
 "Report-Msgid-Bugs-To: pete@akeo.ie\n"
 "POT-Creation-Date: 2020-04-17 12:17+0100\n"
-"PO-Revision-Date: 2020-04-17 12:30+0100\n"
+"PO-Revision-Date: 2020-10-15 07:47+0800\n"
 "Language: zh_CN\n"
 "MIME-Version: 1.0\n"
 "Content-Type: text/plain; charset=UTF-8\n"
@@ -13,7 +13,7 @@ msgstr ""
 "X-Rufus-LCID: 0x0804, 0x1004\n"
 "Last-Translator: \n"
 "Language-Team: \n"
-"X-Generator: Poedit 2.3\n"
+"X-Generator: Poedit 2.4.1\n"
 
 #. • IDD_DIALOG → IDS_DRIVE_PROPERTIES_TXT
 msgid "Drive Properties"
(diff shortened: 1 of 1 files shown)


res/loc/po/zh-CN.po
```diff
@@ -1,9 +1,9 @@
 msgid ""
 msgstr ""
-"Project-Id-Version: 3.5\n"
+"Project-Id-Version: rufus\n"
```
pbatard, line 3: Please don't change `Project-Id-Version`.

res/loc/po/zh-CN.po
```diff
@@ -288,39 +288,39 @@ msgstr "字节"
 #.
 #. *Short* version of the kilobyte size suffix
 msgid "KB"
-msgstr ""
+msgstr "KB"
```
pbatard, line 291: This is unneeded. If a translation is the same as English, then English will be used. Which means that, if "KB" is not translated for Chinese, then it will be correctly be displayed as "KB".

res/loc/po/zh-CN.po
```diff
@@ -288,39 +288,39 @@ msgstr "字节"
...
 
 #. • MSG_027
 msgid "kilobytes"
-msgstr "K 字节"
+msgstr "k 字节"
 
 #. • MSG_028
 msgid "megabytes"
-msgstr "M 字节"
+msgstr "兆 字节"
```
pbatard, line 323: I'm not sure I understand why kilobytes became "k 字节" but megabytes becomes "兆 字节". The rule of whether an English letter should be used before "字节" doesn't seem that logical. Also I would expect that a capital letter would be used, since it derives from English and a capital letter is always being used for KB or MB. All in all, I'm not sure why this change is needed.

res/loc/po/zh-CN.po
```diff
@@ -815,7 +815,7 @@ msgstr "标准 Windows 安装"
 #. http://en.wikipedia.org/wiki/Windows_To_Go in your language.
 #. Otherwise, you may add a parenthesis eg. "Windows To Go (<hint at what it does>)"
 msgid "Windows To Go"
-msgstr ""
+msgstr "转到窗口"
```
pbatard, line 818: My understanding is that "Windows To Go" is not really something you want to translate, as it's pretty much universally known as the English designation. I fear that by not calling it "Windows To Go", people will be confused as to what it is.

Especially, a google search for "转到窗口" doesn't seem to return the hits I would expect about the Windows To Go functionality of Windows.

All in all then, I don't think this change is required. Left untranslated, "Windows To Go" will be used, which is what we probably want.

pbatard (commented):
All in all, I'm going to reject this Pull Request, because I don't currently see much of anything that improves the existing Chinese translation. But thanks for submitting it.

PR #1665, opened 2020-12-17
Update file.h

Doncuppjr:
Increase MAX_DATA_LEN to accommodate larger boot loaders.

### DIFF
--- a/src/ms-sys/inc/file.h
+++ b/src/ms-sys/inc/file.h
@@ -4,7 +4,7 @@
 #include <stdint.h>
 
 /* Max valid value of uiLen for contains_data */
-#define MAX_DATA_LEN 65536
+#define MAX_DATA_LEN 262144
 
 /* We hijack the FILE structure for our own needs */
 typedef struct {


Doncuppjr:
Why so small? Can we let it be the whole size of the MBR gap?

pbatard:
> Why so small? Can we let it be the whole size of the MBR gap?

Because this is tailored for my actual needs in terms of boot loader size, and now what other people might do.

Sorry but unless I happen to need to increase this size for the bootloaders used with the standard version of Rufus, I'm going to reject this request and expect people who want to use the Rufus codebase with custom bootloaders to modify the required elements of code on their own.

Please bear in mind that any application is developed for its current expectations, and not the way people may stretch it to do things that are out of scope, such as embedding large boot loaders.

In other words, as long as I see no case for the embedding of a bootloader larger than 64k, in the vanilla version of Rufus, I'm not going to change this limit (precisely because it's going to tell me if the embedded files start getting too large).

And if you are unhappy about this, please bear in mind that, when I needed more than 10 KB here, rather than ask the ms-sys author to increase the constant in their codebase (see https://sourceforge.net/p/ms-sys/code2/HEAD/tree/trunk/inc/file.h#l5), I just accepted that, as someone who was creating a derivative work, it was my own responsibility to alter the constants, in my own work, as required.

&rarr; Rejected.

Doncuppjr:
No worries. I got to a point that I can work with. It just seemed like an arbitrary constraint vs a real limitation of the 1st partition start byte - 64. Thank you again for all of your hard work.

PR #1715, opened 2021-04-08
Windows PE resource info - add Rufus description

wesinator:
Use "The Reliable USB Formatting Utility" for description in Windows PE
binary builds

**p.s.** thanks for all the work you've put into this great tool ! 
I still haven't found anything else like it that writes non-destructive ISO images, even on Linux.

Cheers,

### DIFF
--- a/res/appstore/Package.appxmanifest
+++ b/res/appstore/Package.appxmanifest
@@ -11,7 +11,7 @@
   <Identity
     Name="19453.net.Rufus"
     Publisher="CN=7AC86D13-3E5A-491A-ADD5-80095C212740"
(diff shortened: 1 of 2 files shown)


pbatard:
I'm going to have to reject this, on account that the description only makes sense for English speakers, and Rufus may be used by people who don't speak English at all.

Especially, the `.appxmanifest` description may end up being displayed in the start menu, so I'd rather have the application name than a description that non-English speaker may not understand...

I appreciate your willingness to try to improve Rufus though. If Rufus was English-only, I would definitely have applied your request!

PR #1716, opened 2021-04-10
Update badblocks.c

GardeningTool:


pbatard:
Unless the [source this code is derived from](https://git.kernel.org/pub/scm/fs/ext2/e2fsprogs.git/tree/lib/ext2fs/badblocks.c#n198) is updated, I'm not planning to pick these kind of improvements, as the less a derivative source deviates from the original, the easier it is to integrate updates.

Please be very mindful when you propose a source change that the code may not be something we wrote, but something we picked from an upstream project, and therefore that needs to be patched there for us to integrate it.

Note that it doesn't mean that I wouldn't prefer the code to be improved in the manner you suggest, just that, sometime, it makes more sense to go with whatever idiosyncrasies exist in the imported code, especially if altering that code won't bring any functional change to the table.

&rarr; Rejected

PR #1721, opened 2021-04-25
Turkish translation update

kayazeren:
Turkish translation completely reviewed and refined
As a translator reference you can check my Transifex profile: https://www.transifex.com/user/profile/kayazeren/


pbatard:
I'm afraid that I don't accept translation updates unless they have been vetted by the active translator, and there is an active translator for Turkish (Mehmetali), whose contact details you can find out in the About dialog when Rufus is running in Turkish.

This may sound counter intuitive (After all, why would a developer refuse improvements?), but we've had issue in the past where translation updates were designed around a political agenda, where a translator was trying to favour a specific dialect or promote "revised" grammatical views that were far from being agreed by the mainstream language speakers, and as such, since I cannot judge the impartiality of a translation update, I have to request that changes, that aren't limited to obvious typos, are done in consultation with the active translator.

That's actually the reason why I'm trying to make sure that you can find the contact details on a translator in the About dialog, so that, if you believe that a translation needs improvement, you do discuss those with that person. Otherwise, you are mostly trying to carry out changes to their work behind their back, which, I hope, you can understand is something we want to avoid.

As a result, I will close this PR. The Turkish translator is still active (he very recently sent a translation for the 3.14 release), so if you feel that the translation needs improvement, I will kindly request that you liaise with them first.

kayazeren:
Thank you for your email.
I understand very well that the translation/translator issues seem easy but
could be very complicated.
After I checked translation quality, I was not sure that
previous translators were still active.
I will contact the active maintainer as you suggested.

Regards
Kaya

Pete Batard ***@***.***>, 25 Nis 2021 Paz, 14:21 tarihinde
şunu yazdı:

PR #1815, opened 2021-10-30
Off-by-one error

mjunix:


pbatard:
I'm going to have to reject this Pull Request because this code does not originate from Rufus but from libcdio.

Therefore, if you want a potential bugfix to be applied there, it should be sent to the [libcdio project](https://www.gnu.org/software/libcdio/) (development mailing list [here](https://lists.gnu.org/mailman/listinfo/libcdio-devel)), so that it can be reviewed and integrated.

It's only if this change is applied upstream that it will naturally find its way into Rufus. But not the other way around.

The reason why I am asking you to do that is because it is much more beneficial __for users__ to have bugs actually fixed in the original upstream project rather than in any downstream project.

Also, whereas you may think that a developer like myself should have the time to just apply a fix like this in their code, and then send a patch upstream, you would be wrong. I'm afraid that my time is limited and it's always better to have the person who found a bug and proposes a fix to champion then with the upstream project.

So please send this fix to libcdio, thank you.

PR #1818, opened 2021-11-01
Fix LGTM.com warning: Unnecessary pass

DimitriPapadopoulos:
https://lgtm.com/projects/g/pbatard/rufus/?severity=warning


pbatard:
Not gonna bother with this, as this python script is no longer useful no that we have switched to using `.po` files and I am planning to remove it eventually.

PR #1845, opened 2021-12-25
[efi] use names as defined in spec

concatime:
Hi.
I renamed the entries as defined in [the latest spec](https://uefi.org/sites/default/files/resources/UEFI_Spec_2_9_2021_03_18.pdf) (`3.5.1.1 Removable Media Boot Behavior`) to stay consistent. This change should alter nothing because FAT is case insensitive.
This change has been been tested yet. I would like your opinion first.
Thank you.

### DIFF
--- a/src/iso.c
+++ b/src/iso.c
@@ -92,10 +92,13 @@ static const char* ldlinux_name = "ldlinux.sys";
 static const char* ldlinux_c32 = "ldlinux.c32";
 static const char* md5sum_name[] = { "MD5SUMS", "md5sum.txt" };
 static const char* casper_dirname = "/casper";
-static const char* efi_dirname = "/efi/boot";
+/* https://uefi.org/sites/default/files/resources/UEFI_Spec_2_9_2021_03_18.pdf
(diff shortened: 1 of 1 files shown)


pbatard:
I appreciate that but we take care of care sensitivity within the code (you should see that we use `stricmp()` and `toupper()` where needed) and, considering that the code that uses these definitions is __not__ targeted at FAT, but at other file systems where case sensitivity applies (e.g. EFI bootloaders lookup on ISO9660 with Joliet, UDF, etc.), there isn't much point in following the specs, since we very much want to be able to handle the many existing cases where the casing of the bootloaders isn't going to match the specs.

So it's actually more advantageous to deliberately deviate from the specs in terms of casing, when we declare the bootloader names, so that, if there's any issue with the code we explicitly designed for handling of bootloader names regardless of casing, we'll pick on that before someone complains that their potentially non specs-compliant bootable image isn't detected as EFI bootable...

As such, I don't think applying this change is worth it. But I hope that won't dissuade you from proposing other Pull Requests, if you think there are other improvements that could be made.

PR #1863, opened 2022-01-30
[loc] update MSG_235, MSG_191 & MSG_192

stabacov:
Japanese translation update.

It was looks like "Bad blocks: The write passed. 1/1 - ...".
My suggestion is like "Bad blocks: Writing 1/1 pass - ...".


pbatard:
Applied. Thanks!

PR #1888, opened 2022-03-18
Fixed minor bugs and added CRT support for multithreaded app in MSVC

GermanAizek:
@pbatard check this commit and write feedback please.

### DIFF
--- a/src/checksum.c
+++ b/src/checksum.c
@@ -1134,7 +1134,11 @@ DWORD WINAPI SumThread(void* param)
 	int num_checksums = CHECKSUM_MAX - (enable_extra_hashes ? 0 : 1);
 
 	if ((image_path == NULL) || (thread_affinity == NULL))
+#if _MSC_VER && !__INTEL_COMPILER
+		_endthreadex(r);
+#else
 		ExitThread(r);
+#endif
 
 	uprintf("\r\nComputing checksum for '%s'...", image_path);
 
@@ -1155,7 +1159,11 @@ DWORD WINAPI SumThread(void* param)
 			uprintf("Unable to create checksum thread event: %s", WindowsErrorString());
 			goto out;
 		}
+#if _MSC_VER && !__INTEL_COMPILER
+		sum_thread[i] = (HANDLE)_beginthreadex(NULL, 0, &IndividualSumThread, (LPVOID)(uintptr_t)i, 0, NULL);
+#else
 		sum_thread[i] = CreateThread(NULL, 0, IndividualSumThread, (LPVOID)(uintptr_t)i, 0, NULL);
+#endif
 		if (sum_thread[i] == NULL) {
 			uprintf("Unable to start checksum thread #%d", i);
 			goto out;
@@ -1250,7 +1258,11 @@ out:
 	PostMessage(hMainDialog, UM_FORMAT_COMPLETED, (WPARAM)FALSE, 0);
 	if (r == 0)
 		MyDialogBox(hMainInstance, IDD_CHECKSUM, hMainDialog, ChecksumCallback);
+#if _MSC_VER && !__INTEL_COMPILER
+	_endthreadex(r);
+#else
 	ExitThread(r);
+#endif
 }
 
 /*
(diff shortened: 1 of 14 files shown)


src/libcdio/cdio/ecma_167.h
```diff
@@ -334,6 +334,7 @@ struct anchor_vol_desc_ptr_s
 } GNUC_PACKED;
 
 typedef struct anchor_vol_desc_ptr_s anchor_vol_desc_ptr_t;
+typedef struct anchor_vol_desc_ptr_s *ptr_anchor_vol_desc_ptr_t;
```
pbatard, line 337: I'm afraid I will not patch any of the src/libcdio/** files unless they are patched in the upstream project (https://www.gnu.org/software/libcdio/), so, if you will need to send this part of your patch (based on the actual libcdio code and not the Rufus version) to the libcdio-devel mailing list at https://savannah.gnu.org/mail/?group=libcdio.

src/libcdio/cdio/iso9660.h
```diff
@@ -391,6 +391,7 @@ struct iso9660_pvd_s {
 } GNUC_PACKED;
 
 typedef struct iso9660_pvd_s  iso9660_pvd_t;
+typedef struct iso9660_pvd_s  *ptr_iso9660_pvd_t;
```
pbatard, line 394: See above.

src/libcdio/iso9660/iso9660_fs.c
```diff
@@ -368,7 +368,7 @@ get_member_id(iso9660_t *p_iso, cdio_utf8_t **p_psz_member_id,
     strip = false;
     (*p_psz_member_id)[j] = pvd_member[j];
   }
-  if (strlen(*p_psz_member_id) == 0) {
```
pbatard, line 371: See above.

src/net.c
```diff
@@ -54,7 +54,7 @@ extern loc_cmd* selected_locale;
 extern HANDLE dialog_handle;
 extern BOOL is_x86_32;
 static DWORD error_code, fido_len = 0;
-static BOOL force_update_check = FALSE;
+volatile static BOOL force_update_check = FALSE;
```
pbatard, line 57: My understanding is that volatile is for variables that can be altered outside of the code for the application (e.g. from an interrupt routine that would not be part of Rufus). So I'm not sure I understand why it should be needed here. Even if it is to be changed from another thread, the compiler is smart enough not to need volatile and a quick search for "is volatile needed for multithreaded" will show that it is unnecessary for that purpose. Therefore I have to ask: Have you experienced a specific issue here that makes you believe that volatile is needed?

src/checksum.c
```diff
@@ -1134,7 +1134,11 @@ DWORD WINAPI SumThread(void* param)
 	int num_checksums = CHECKSUM_MAX - (enable_extra_hashes ? 0 : 1);
 
 	if ((image_path == NULL) || (thread_affinity == NULL))
+#if _MSC_VER && !__INTEL_COMPILER
```
pbatard, line 1137: I'd prefer to have this as a global `EXITTHREAD()` macro defined in `rufus.h` so that we don't have to use that much `#if` boilerplate in the code. Also, my convention for `#if` in Rufus is to go `#if defined(_MSC_VER) && !defined(__INTEL_COMPILER)`, with explicit `define(...)` so please try to follow the same, thank you.

src/checksum.c
```diff
@@ -1155,7 +1159,11 @@ DWORD WINAPI SumThread(void* param)
 			uprintf("Unable to create checksum thread event: %s", WindowsErrorString());
 			goto out;
 		}
+#if _MSC_VER && !__INTEL_COMPILER
```
pbatard, line 1162: Same remark as above, with a `CREATETHREAD()` macro. Now of course the issue here is that `IndividualSumThread` is dereferenced in one case and not in the other, which I'm not too sure how to sort out in a macro, so we might have to go with an `static __inline` function where we pass `&IndividualSumThread` always and just use `*` of that param for the `CreateThread` version.

Note that if you go for a `static __inline` call here, then I'd prefer you did the same for both calls and used something `BeginThread()` and `EndThread()` for the names.

pbatard:
Thanks. Please see my remarks inline. I don't have an objection for the thread stuff (but can you please describe the issue you ran into with threading in more details?), but some of the libcdio stuff needs to be patched in the upstream project, so you may want to break this into 2 PRs, so that we can apply the threading stuff sooner.

PR #1929, opened 2022-05-16
Visitors Badge

itsmeshibintmz:
Added visitors badge to repository readme.

### DIFF
--- a/README.md
+++ b/README.md
@@ -2,6 +2,7 @@ Rufus: The Reliable USB Formatting Utility
 ==========================================
 
 ![Badge](https://visitor-counter-badge.vercel.app/api/pbatard/rufus)
+
 [![VS2022 Build Status](https://img.shields.io/github/workflow/status/pbatard/rufus/VS2022.svg?style=flat-square&label=VS2022%20Build)](https://github.com/pbatard/rufus/actions/workflows/vs2022.yml)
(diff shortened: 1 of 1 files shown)


pbatard:
I'm sorry but whereas I trust https://shields.io because its inception & mandate are pretty clear (and they were clearly designed in the spirit of Open Source and not for profit), I have a hard time doing the same for a subdomain of https://vercel.app, because I'm going to assume that, whereas https://shields.io makes clear that they aren't in it for tracking, your visitor counter may log the IP of every visitor that accesses the project, and do whatever it wants with that data.

If I were to add a counter for stats that (unlike the downloads) cannot be publicly accessed from GitHub, then I want to be the one who controls how the data is collected and what is done with it, so that I can reassure visitors (who choose to trust the developer of the software) that they are not being tracked beyond the obvious and only goal of incrementing a counter.

As such, I'm afraid I'm going to have to reject this Pull Request because I simply can't trust a random user who wants to incorporate remote content from a server that seems a bit too obscure (to me) to be able to vet as trustworthy. There are just too many ways in which blindly trusting someone with remote content may result in nefarious behaviour for visitors of this project...

&rarr; Rejected

itsmeshibintmz:
Okay understood.

PR #1930, opened 2022-05-17
Updated visitors count badge

itsmeshibintmz:
used shields io visitor counter.

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,7 +1,7 @@
 Rufus: The Reliable USB Formatting Utility
 ==========================================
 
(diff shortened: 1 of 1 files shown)


pbatard:
Sorry but https://shields-io-visitor-counter.herokuapp.com is not a shields.io URL. It's a herokuapp.com URL, so the same issue of trust as #1929 applies.

But, to cut this whole thing short, even if it was provided by shields.io, I don't feel like the README for the project needs a visitor badge, so I'm not going to accept a Pull Request to add a visitor badge anyway.

PR #1948, opened 2022-06-07
Update pt-BR.po

marcosfrm:
Small fixes.


pbatard:
Thanks Marco. I have not applied these changes for the BETA but I'll apply them for the 3.19 release. The reason these are not applied to the BETA is because, _if you want_, there are [a few additional entries I have added](https://github.com/pbatard/rufus/blob/master/res/loc/ChangeLog.txt#L8-L19), that you may also want to include (knowing that, you don't really have to do that, as I'll run a translation round with all the translators for the 3.20 release, and it's possible I'm going to alter these new messages further over the coming months).

marcosfrm:
If other translators do not show up, I will wait for the 3.20 call.

PR #2010, opened 2022-08-04
Update ko-KR.po

VenusGirl:
Korean Update


pbatard:
Thanks but I'm afraid I don't take application updates until the next translation round (which hasn't started because I am still figuring out additional messages that will need to be translated).

I will therefore put this PR on hold until I am ready to process it, and refer it to the active Korean translator for review.

VenusGirl:
It was translated accurately and perfectly because there was an error in the existing translation and the user could malfunction.

PR #2022, opened 2022-08-15
Update ko-KR

VenusGirl:


pbatard:
Again I will not take application updates until the next translation round starts, and it has not started yet because I am still figuring out additional messages that will need to be translated.

Therefore I will put this request on hold yet again,

PR #2024, opened 2022-08-17
Update Japanese translation

s1204IT:
I already understand that we are withholding PR.
It does not matter if it is late.


pbatard:
I'm going to defer this, just like the other translation update PRs.

PR #2104, opened 2022-12-05
Use strtoull instead of atoll

AtariDreams:
This means we don't need to worry about conversion issues regarding signedess. In addition, the behavior will no longer be undefined if for some reason the conversion cannot happen.


pbatard:
Applied. Thanks!

PR #2165, opened 2023-02-13
Update rufus.loc

VenusGirl:
Korean Update


pbatard:
I don't accept Pull Request with `rufus.loc` updates. And as mentioned when opening a Pull Request, I don't accept unsolicited translation updates because it creates a massive headaches. For instance, what if someone happened to submit an update for Korean before you do. Should I just tell you you did a lot of work for nothing because whoever comes first "wins"?

I will deal with translation updates when I do start a translation round, and a translation round starts with me first notifying the existing translators by e-mail, and then, if the translators do not provide their translation in a timely manner, looking beyond the scope of the existing translators. And whereas I do appreciate the desire to help, I have to point out that it's just too much of a hassle as well as way too time consuming to have to deal with unsolicited requests, especially as my understanding is that [there are some Korean translators out there that are motivated with trying to push a political agenda](https://github.com/pbatard/rufus-web/issues/8) and therefore, as I am not a Korean speaker, I will favour Korean translators I have already worked with over ones that seem a bit too eager to see their translation changes integrated...

PR #2186, opened 2023-02-24
Update rufus.loc

VenusGirl:
Add ko-kr 321 - 336
Add ko-kr 901 - 922


pbatard:
Since the existing Korean translator have not manifested themselves, I have now applied your translation update (from the e-mail you sent on 2023.02.12, and not from this `.loc` update since I can only work with a `.po` file and your `.loc` changes).

Thanks for working on this!

PR #2289, opened 2023-07-24
Volume label does not a directory anymore

AJIOB:
Fix issue #2288 

As [wikipedia](https://en.wikipedia.org/wiki/Design_of_the_FAT_file_system#Directory_entry) is written for directory entries, file attribute with value `0x08` is a volume label.

# Rufus log after patch with problematic ISO

```
Rufus x64 v4.2.2072
Windows version: Windows 10 Pro x64 (Build 19045.3208)
Grub versions: 0.4.6a, 2.06
Localization set to 'en-US'
Notice: The ISO download feature has been deactivated because 'Check for updates' is disabled in your settings.
Found 517 officially revoked UEFI bootloaders from embedded list
Found 2351 additional revoked UEFI bootloaders from this system's SKUSiPolicy.p7b
Found USB 2.0 device 'JetFlash Transcend 8GB USB Device' (8564:1000)
NOTE: This device is a USB 3.0 device operating at lower speed...
1 device found
Disk type: Removable, Disk size: 7.9GB, Sector size: 512 bytes
Cylinders: 960, Tracks per cylinder: 255, Sectors per track: 63
Partition type: SFD (Super Floppy Disk) or unpartitioned

Image provided: 'D:\Projects\my-iso.iso'
Scanning image...
ISO analysis:
  Image is an ISO9660 image
  Detected EFI bootloader(s) (from '/boot/grub/efiboot.img'):
  ● 'bootx64.efi'
Disk image analysis:
  Image does not have a Boot Marker
ISO label: 'MyIsoImage'
  Size: 497 MB (Projected)
  Uses: EFI
Using image: my-iso.iso (497.4 MB)
```

### DIFF
--- a/src/syslinux/libfat/searchdir.c
+++ b/src/syslinux/libfat/searchdir.c
@@ -40,7 +40,8 @@ int32_t libfat_searchdir(struct libfat_filesystem *fs, int32_t dirclust,
 
 	for (nent = 0; nent < LIBFAT_SECTOR_SIZE;
 	     nent += sizeof(struct fat_dirent)) {
-	    if (!memcmp(dep->name, name, 11)) {
+	    // Volume label entry must be ignored: it is not a real directory
+	    if ((dep->attribute & 0x08) == 0 && !memcmp(dep->name, name, 11)) {
 		if (direntry) {
 		    memcpy(direntry->entry, dep, sizeof(*dep));
 		    direntry->sector = s;


pbatard:
Thanks for the patch.

If I understand correctly, you created a `efiboot.img` with volume label `EFI`, which got translated to a FAT directory entry for a volume label (File Attribute `0x08`). Therefore, since there were 2 file directory entries for `EFI` (1 for the volume label, with attribute `0x08` and one for the subdirectory, probably with attribute `0x10`), the SysLinux FAT `searchdir()` code interpreted the first one as the directory it was looking for, and obviously could therefore not locate the UEFI bootloaders.

It does indeed make sense to want to to filter out volume labels, so I will apply your patch when I get a chance. I also think that, to be consistent, we want to filter out entries with the `Device` attribute (`0x40`) as well, as I could technically see someone creating a `EFI` device entry just for the heck of it, which would also collide with the search for the `EFI` directory.

pbatard:
This should be fixed now, so thanks again for the patch.

Feel free to test with the artifacts from https://github.com/pbatard/rufus/actions/runs/5650627386 (the latest build that includes the patch) and don't hesitate to let me know if you still see the problem.

AJIOB:
Thanks, @pbatard.

Your patch works!

PR #2414, opened 2024-01-28
add circle linux as derivatives rhel

bella485:
Circle Linux published 8.x and 9.x ,  some user and developer feed back problems used rufus make USB Bootable.


pbatard:
Applied - thanks!

Issue #2451, opened 2024-03-26
Add detection for unattend.xml file


daboynb:
﻿Checklist
---------
- [x] I looked at https://github.com/pbatard/rufus/wiki/FAQ to see if my question has already been answered.
- [x] I performed a search in the issue tracker for similar issues using keywords relevant to my problem, such as the error message I got from the log.

Issue description
-----------------
When a user provides an ISO file containing an unattend.xml file and uses Rufus, the software still prompts for unattended options. 

This behavior will lead to this error 
![Capture](https://github.com/pbatard/rufus/assets/106079917/76982e71-a450-4670-a874-441262011a6f)

To address this issue, Rufus should include a check to determine if the unattend.xml file is already present. 

If the file is found, Rufus should skip prompting for unattended options.

daboynb:
Here the log, sorry.
[rufus.log](https://github.com/pbatard/rufus/files/14764262/rufus.log)

PR #2458, opened 2024-04-09
Fix for https://github.com/pbatard/rufus/issues/2451

daboynb:
Check if unattend.xml exist, if yes, exit. Do not apply any windows customization.


pbatard:
Thanks but I think I'd rather have the check earlier during ISO scan so that Windows customisation is not being proposed at all, which will provide a better hint to the user that there is an `unattend.xml` on the source image than an easy to miss message in the log.

daboynb:
Okay, thank you.

Issue #2500, opened 2024-06-12
missing: line 81: aclocal-1.14: command not found


Marietto2008:
Hello.

In relation to this fork :

https://github.com/markpeek/rufus/tree/markpeek-bhyve

that wants that Rufus detects an USB stick when it is passed as a bhyve virtual disk, I tried to compile that code,but I'm having a problem to compile it in my Ubuntu (23.10) installation :

```
root@Z390-AORUS-PRO-DEST:/home/ziomario/Scaricati/rufus# make

CDPATH="${ZSH_VERSION+.}:" && cd . && /bin/bash /home/ziomario/Scaricati/rufus/missing aclocal-1.14 
/home/ziomario/Scaricati/rufus/missing: line 81: aclocal-1.14: command not found
WARNING: 'aclocal-1.14' is missing on your system.
         You should only need it if you modified 'acinclude.m4' or
         'configure.ac' or m4 files included by 'configure.ac'.
         The 'aclocal' program is part of the GNU Automake package:
         <http://www.gnu.org/software/automake>
         It also requires GNU Autoconf, GNU m4 and Perl in order to run:
         <http://www.gnu.org/software/autoconf>
         <http://www.gnu.org/software/m4/>
         <http://www.perl.org/>
make: *** [Makefile:277: aclocal.m4] Error 127

root@Z390-AORUS-PRO-DEST:/home/ziomario/Scaricati/rufus# apt reinstall aclocal-1.14
Reading package lists... Done
Building dependency tree... Done
Reading state information... Done
E: Unable to locate package aclocal-1.14
E: Couldn't find any package by glob 'aclocal-1.14'

root@Z390-AORUS-PRO-DEST:/home/ziomario/Scaricati/rufus# apt search aclocal
Sorting... Done
Full Text Search... Done

root@Z390-AORUS-PRO-DEST:/home/ziomario/Scaricati/rufus# automake --version
automake (GNU automake) 1.16.5
Copyright (C) 2021 Free Software Foundation, Inc.
License GPLv2+: GNU GPL version 2 or later <https://gnu.org/licenses/gpl-2.0.html>
This is free software: you are free to change and redistribute it.
There is NO WARRANTY, to the extent permitted by law.

Written by Tom Tromey <tromey@redhat.com>
       and Alexandre Duret-Lutz <adl@gnu.org>.
```

I prefer to compile it under Linux than under Windows,where I have 0 experience.

thecatontheceiling:
[...]

PR #2504, opened 2024-06-14
[vhd] allow the use of FreeBSD bhyve virtual disks

markpeek:
FreeBSD bhyve virtualization does not allow passthrough of individual USB devices. This change allows the detection of the bhyve virtual device volume similar to how VMware and HyperV is handled.

Verified by @marietto2008 via https://github.com/pbatard/rufus/issues/2500#issuecomment-2166982401 and https://bugs.freebsd.org/bugzilla/show_bug.cgi?id=279659#c11

### DIFF
--- a/src/dev.c
+++ b/src/dev.c
@@ -287,6 +287,7 @@ static __inline BOOL IsVHD(const char* buffer)
 		"Arsenal_________Virtual_",
 		"KernSafeVirtual_________",
 		"Msft____Virtual_Disk____",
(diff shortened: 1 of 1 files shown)


pbatard:
Thanks. This looks straightforward and low-impact enough, so it will be added to the next version of Rufus.

markpeek:
Thank you.

pbatard:
Applied. Thanks again!

PR #2539, opened 2024-08-13
libcdio: fix MSVC-only definition

ognevny:
according to the comment it should be defined for MSVC only. fixes UCRT build

### DIFF
--- a/src/libcdio/config.h
+++ b/src/libcdio/config.h
@@ -40,7 +40,9 @@
 /* Define to 1 if you have the `fseeko64' function. */
 #define HAVE_FSEEKO64 1
 /* The equivalent of fseeko64 for MSVC is _fseeki64 */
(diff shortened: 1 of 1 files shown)


ognevny:
signed commit is required... ok, will push it from github directly

pbatard:
Signed commit is not required for a PR. Proof is it was created and appeared as a Rufus PR, which is fine by me, regardless of whether it's signed or not.

ognevny:
I've already reopened #2540

pbatard:
Yeah. I guess it's needed by GitHub if you wanna run Actions, but if Actions fail when I push the PR, I'm gonna fix it anyway (though of course, I expect you to have cloned the repo and tested with Actions before created the PR first).

ognevny:
in this case I'm _sure_ that it works on MinGW

PR #2540, opened 2024-08-13
libcdio: fix MSVC-only definition

ognevny:
according to the comment it should be defined for MSVC only. fixes UCRT MinGW build


ognevny:
@pbatard

pbatard:
Yeah, yeah. It will happen when it happens. I'm working on other stuff right now and I'm afraid it's hard to see this as high priority, even if it's a simple PR.

PR #2550, opened 2024-08-23
mbr.S usb boot function update

PrIzRaKDev:
Decided to update the usb boot function as there were some aspects to rewrite. Here's what I changed:

Consistent indents and spacing
Removed unnecessary comments and added descriptive labels Used `.label` notation for local labels instead of `label:` notation Removed the `ds:` prefix from memory references, as it is not needed in 16-bit code Used square brackets `[]` for memory references instead of parentheses `()` Improved code organization and readability
Removed unnecessary `mov` instructions and optimized code flow.

I tried not to change the function's logic

### DIFF
--- a/res/mbr/mbr.S
+++ b/res/mbr/mbr.S
@@ -151,22 +151,29 @@ boot_drive:
 
 # ---------------------------------------------------------------------------
 
-boot_usb_rem_rtc:		# Boot USB drive (0x80)
-	call restore_rtc_vect	# Remove our RTC override
-boot_usb:
-	call flush_keyboard	# Make sure the keyboard buffer is clear
-	mov  bx, offset partition_table
-	mov  dx, ds:[bx]
-	push dx
-	mov  dl, 0x80		# Override disk number, as we're not using our int yet
-	mov  cx, ds:[bx+2]
-	mov  ebx, ds:[bx+8]	# Must come last since it modifies BX
-	call read_sector
-	jnb  boot_drive
-	pop  es			# failed to read PBR from USB - exit back to BIOS
-	pop  ds
-	retf
 
+boot_usb_rem_rtc:
+    call restore_rtc_vect  
+
+boot_usb:
+    call flush_keyboard     
+
+    mov bx, offset partition_table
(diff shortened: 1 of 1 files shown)


pbatard:
I don't quite like the comment removal. Because this is assembly as well as (potentially) something that might be used by other as a reference, the whole point of these comments is to be super descriptive about what we're doing. But suddenly you remove comments on a whole section because, *according to you* they are unnecessary...

You also state that you changed labels to dot notation and use parenthesis instead of square brackets, whereas this is not done in your patch (and I don't think I want that anyway, as I designed the assembly for the assembler and syntax that I prefer, and I am not going change that, at the risk of breaking compatibility with the assembler I actually use, because someone has a different preference. I'm not here to prettify code according to someone else's tastes, sorry). And your "consistent indents and spacing" actually introduces indents issue, since you suddenly replace tabs, which is what the whole file uses for indent, with spaces on the section you modified.

You also, very importantly, do not explain why you suddenly want to do that? What exactly are you trying to solve here? Are you trying to use an assembler where column labels and square brackets don't work?

The only actual actionable element I see from your patch is that, whereas we should have used the `ds:` prefix in a couple "naked" square bracket references, to be both consistent but more importantly very explicit, we did forget to do so in a couple of lines. But this will do the exact opposite of what you wanted.

Now, I'm going to assume that you're relatively new to creating pull request, so please don't let the fact that I am going to ultimately reject this patch deter you. We have all had to start somewhere and I don't know of any developer who actually gets annoyed at someone who is only ultimately trying to improve a project's codebase. Just bear in mind that your patch description needs to match what the patch actually does, that your patch should avoid introduce whitespaces issues, and that patches that boil down to a _"but I like it better when the code is formatted this way"_ don't generally gain much traction, as you are instead supposed to stick with whatever formatting preferences the original developer(s) used, regardless of whether you like them or not, and only introduce formatting preference patches if it's to fix internal consistency.

&rarr; Rejected, even though I will try to fix the two `[bx]` that are missing a segment prefix, and will try to credit you in the commit message when I do (but will not create an actual commit for this change).

PR #2589, opened 2024-10-17
[iso] Switch the ReactOS path to freeldr.sys - Allow ReactOS to boot from a flash rufus usb again

DarkFire01:
https://github.com/reactos/reactos/commit/764881a94b4129538d62fda2c99cfcd1ad518ce5
Follow this commit rufus no longer works with ReactOS's newer ISOs. This PR just aims to quickly fix this by checking for both setupldr.sys and freeldr.sys.

### DIFF
--- a/src/iso.c
+++ b/src/iso.c
@@ -112,7 +112,8 @@ static const char* syslinux_cfg[] = { "isolinux.cfg", "syslinux.cfg", "extlinux.
 static const char* isolinux_bin[] = { "isolinux.bin", "boot.bin" };
 static const char* pe_dirname[] = { "/i386", "/amd64", "/minint" };
 static const char* pe_file[] = { "ntdetect.com", "setupldr.bin", "txtsetup.sif" };
-static const char* reactos_name = "setupldr.sys"; // TODO: freeldr.sys doesn't seem to work
+static const char* reactos_name = "freeldr.sys";
+static const char* reactos_old_name = "setupldr.sys";
 static const char* kolibri_name = "kolibri.img";
 static const char* autorun_name = "autorun.inf";
 static const char* manjaro_marker = ".miso";
@@ -274,9 +275,12 @@ static BOOL check_iso_props(const char* psz_dirname, int64_t file_length, const 
 			}
 		}
 
-		// Check for ReactOS' setupldr.sys anywhere
-		if ((img_report.reactos_path[0] == 0) && (safe_stricmp(psz_basename, reactos_name) == 0))
-			static_strcpy(img_report.reactos_path, psz_fullpath);
+		// Check for ReactOS presence
+		if (img_report.reactos_path[0] == 0) {
+			if ((safe_stricmp(psz_basename, reactos_name) == 0) ||
+				(safe_stricmp(psz_basename, reactos_old_name) == 0))
+					static_strcpy(img_report.reactos_path, psz_fullpath);
+		}
 
 		// Check for the first 'efi*.img' we can find (that hopefully contains EFI boot files)
 		if (!HAS_EFI_IMG(img_report) && (safe_strlen(psz_basename) >= 7) &&


src/iso.c
```diff
@@ -274,9 +275,13 @@ static BOOL check_iso_props(const char* psz_dirname, int64_t file_length, const
 			}
 		}
 
-		// Check for ReactOS' setupldr.sys anywhere
-		if ((img_report.reactos_path[0] == 0) && (safe_stricmp(psz_basename, reactos_name) == 0))
-			static_strcpy(img_report.reactos_path, psz_fullpath);
+		// Check for ReactOS' presence
```
binarymaster, line 278:
```suggestion
		// Check for ReactOS presence
```
Since the comment now doesn't mention ReactOS' files :)

src/iso.c
```diff
@@ -274,9 +275,13 @@ static BOOL check_iso_props(const char* psz_dirname, int64_t file_length, const
...
+		if (img_report.reactos_path[0] == 0)
+		{
+		    if ((safe_stricmp(psz_basename, reactos_name) == 0) ||
+		        (safe_stricmp(psz_basename, reactos_old_name) == 0))
+		        static_strcpy(img_report.reactos_path, psz_fullpath);
```
binarymaster, line 283:
```suggestion
			if ((safe_stricmp(psz_basename, reactos_name) == 0) ||
				(safe_stricmp(psz_basename, reactos_old_name) == 0))
				static_strcpy(img_report.reactos_path, psz_fullpath);
```
remaining spaces to tabs

src/iso.c
```diff
@@ -274,9 +275,13 @@ static BOOL check_iso_props(const char* psz_dirname, int64_t file_length, const
 			}
 		}
 
-		// Check for ReactOS' setupldr.sys anywhere
-		if ((img_report.reactos_path[0] == 0) && (safe_stricmp(psz_basename, reactos_name) == 0))
-			static_strcpy(img_report.reactos_path, psz_fullpath);
+		// Check for ReactOS presence
+		if (img_report.reactos_path[0] == 0)
+		{
```
binarymaster, line 280:
```suggestion
		if (img_report.reactos_path[0] == 0) {
```
adapting to current code style

DarkFire01:
Sorry I had never done commit signing before this so it took a ... few tries to figure out :D

src/iso.c
```diff
@@ -274,9 +275,12 @@ static BOOL check_iso_props(const char* psz_dirname, int64_t file_length, const
 			}
 		}
 
-		// Check for ReactOS' setupldr.sys anywhere
-		if ((img_report.reactos_path[0] == 0) && (safe_stricmp(psz_basename, reactos_name) == 0))
-			static_strcpy(img_report.reactos_path, psz_fullpath);
+		// Check for ReactOS presence
+		if (img_report.reactos_path[0] == 0) {
+			if ((safe_stricmp(psz_basename, reactos_name) == 0) ||
+				(safe_stricmp(psz_basename, reactos_old_name) == 0))
```
HBelusca, line 281: It needs also tabs for the sub-level vertical alignment of the second line of the if-condition?

HBelusca (approved):

binarymaster (approved):

src/iso.c
```diff
@@ -274,9 +275,12 @@ static BOOL check_iso_props(const char* psz_dirname, int64_t file_length, const
...
+					static_strcpy(img_report.reactos_path, psz_fullpath);
```
archeYR, line 282: Since there was a comment suggesting a problem with using freeldr.sys in (probably older?) ReactOS ISOs. Maybe it could make sense to ensure that setupldr.sys is always used whenever it exists, and check for it until it is found or we went through all files on ISO.

```suggestion
		if ((safe_strstr(img_report.reactos_path, reactos_old_name) == NULL) &&
			((safe_stricmp(psz_basename, reactos_name) == 0) ||
			(safe_stricmp(psz_basename, reactos_old_name) == 0)))
			static_strcpy(img_report.reactos_path, psz_fullpath);
```

HBelusca: The problem was that freeldr.sys file was absent from livecd, only setupldr was there, but both files have the same content, so it should be fine ...

Now, I should add, that in much older reactors versions (like the 0.3.x ones), setupldr contained only the startup code to start the bootcd usetup, and it was different from the regular freeldr used in the livecd or in a regular ReactOS installation on a HDD.

pbatard:
@archeYR, I'm afraid I'm not planning to add extra searches that cater specifically for ISOs that are unlikely to be used that commonly, such as ISOs that use both `.sys`.

Also the way I am planning to rework this PR (to be consistent with what I do elsewhere is):
```C
static const char* reactos_name[] = { "setupldr.sys", "freeldr.sys" };

(...)

if (img_report.reactos_path[0] == 0) {
    for (i = 0; i < ARRAYSIZE(reactos_name); i++)
        if (safe_stricmp(psz_basename, reactos_name[i]) == 0)
            static_strcpy(img_report.reactos_path, psz_fullpath);
}
```

This way, if you guys switch the names of the `.sys` files again, it'll be a simple matter of adding a new string to the array.

Also, I guess that means it is actually possible to create bootable USB drives from ReactOS ISOs?
Last time I tried, which I admit was years ago, I wasn't too successful with it, but I decided to add the code anyway, in the hope that USB boot would eventually be sorted out. So it's nice to hear that this has finally happened.

archeYR:
> @archeYR, I'm afraid I'm not planning to add extra searches that cater specifically for ISOs that are unlikely to be used that commonly, such as ISOs that use both `.sys`.

It's okay, I don't care about these old versions that much either.

> Also, I guess that means it is actually possible to create bootable USB drives from ReactOS ISOs?
Last time I tried, which I admit was years ago, I wasn't too successful with it, but I decided to add the code anyway, in the hope that USB boot would eventually be sorted out. So it's nice to hear that this has finally happened.

0.4.15 will include some USB and PnPMgr fixes that allow ReactOS to boot from USB on some real hardware. Also there is a HybridCD build target that has an option to boot ReactOS from RAM disk and it already works reliably.

pbatard:
Applied, thanks! This will be part of the upcoming 4.6 release.

PR #2666, opened 2025-01-22
Added 2 subprojects named Pollock and setup into main project refus

Hurrieam:
I added 2 subprojects named Pollock and setup into the main solution refus so that compiling them in the meantime.

I added the .gitignore file to the solution setup to ignore unnecessary files while compiling them separately.


pbatard:
Sorry but no. It is by design that Pollock and setup are not part of the main solution. And I don't want to add a whole slew of gitignore files that I know I'll never ever need (like _Chutzpah*, _TeamCity or even nuget).

Issue #2576, opened 2024-10-09
[INTERNAL] List of rufus-4.7 enhancements and fixes


pbatard:
﻿﻿﻿﻿**This topic is flagged as INTERNAL and is therefore NOT open for comments.**

### Medium Term

- [x] Add an [_"I've seen a deceptive ad on your website - How could you?!"_ entry to the FAQ](https://github.com/pbatard/rufus/wiki/FAQ#user-content-I_have_seen_some_deceptive_ads_on_your_website_How_dare_you).
- [x] [88f023e79623c081944d5f4ea76a9f64414650c3] Need to update for the FreeDOS 1.4 release. Also Grub4DOS could do with one.
- [x] [fdde687d4681d58b6eaba8e25b1561bea0614eb5] Do we want to remote download the DBXs, like we do SBAT? _Now that Microsoft makes them available from https://github.com/microsoft/secureboot_objects and they have finally relaxed their super-obnoxious license terms, it is more feasible than ever, though downloading about 40 KB on each run isn't ideal. We'd have to store the data between sessions, and check if it's been updated, and prompt the user for the download if it has been_
- [x] [pbatard/rufus-web@2d86633f8be0078a33a974b0d07fd3f8ee02d64b] Parallel to the above, we might need an automated check to lookup and update the SVN value(s) we keep in [`sbat_level.txt`](https://github.com/pbatard/rufus-web/blob/gh-pages/sbat_level.txt) (also, do we need to break down the SVN by arch like Microsoft does?)
- [x] ~~Do we want to open the log automatically in case of error?~~ &rarr; _Not for now_
- [x] Double check that `net accounts` does report `Maximum password age (days): Unlimited` on local account creation &rarr; _Looks good_
- [x] [50801a47ffc947a450740100ba76019519cb7b6a] I am starting to see an awful lot of [This system was unable to provide an English error message](https://preview.redd.it/ca8mdmw8i7xd1.png?width=888&format=png&auto=webp&s=b4593f9c56ee8e07ced1d7de4fd5365fcb4af0f4) (NB: `0xC003045D` is `ERROR_IO_DEVICE` = `1117L`, which we should be able to get in English). Did we regress something with the last error message updates? &rarr; _Apply #2687 and test errors on localized Windows_
[...]

Issue #2524, opened 2024-07-25
'Error: Success' may be reported on faulty drives


pbatard:
Issue description
-----------------

This is a follow up from [this report from reddit](https://old.reddit.com/r/iiiiiiitttttttttttt/comments/1ea54zo/thanks_screw_you_too/) of people managing to get the following error dialog:

![vqtvqkmx39ed1](https://github.com/user-attachments/assets/fb9b2948-527d-4f8a-85df-8fe8f5cce0ac)

Obviously, even if this leads to humorous threads on social media, we'd like to fix this...

Log
---
```
Format operation started
Requesting disk access...
Will use 'E:' as volume mountpoint
Opened \\.\PhysicalDrive2 for shared write access
Analyzing existing boot records...
Drive has a DOS/NT/95A (F2) Master Boot Record
Clearing MBR/PBR/GPT structures...
Erasing 128 sectors
Initializing disk...
Partitioning (GPT)...
● Creating Main Data Partition (offset: 1048576, size: 14.8 GB)
Write error [0x80310000] The operation completed successfully.
Retrying in 5 seconds...
Write error [0x80310000] The operation completed successfully.
Retrying in 5 seconds...
Write error [0x80310000] The operation completed successfully.
Retrying in 5 seconds...
Write error [0x80310000] The operation completed successfully.
Could not zero Main Data Partition: [0xC003001D] The system cannot write to the specified device.
● Creating UEFI:NTFS Partition (offset: 15937253376, size: 1 MB)
Writing UEFI:NTFS data...
Write error [0x80310000] The operation completed successfully.
Retrying in 5 seconds...
Write error [0x80310000] The operation completed successfully.
Retrying in 5 seconds...
Write error [0x80310000] The operation completed successfully.
Retrying in 5 seconds...
Write error [0x80310000] The operation completed successfully.
Write error: [0xC003001D] The system cannot write to the specified device.
\\?\GLOBALROOT\Device\HarddiskVolume29\ was successfully mounted as E:
Re-mounted volume as E: after error
```

PR #2687, opened 2025-02-10
[misc] Fall back to user/system default locale when getting error description

Wack0:
I did some investigation of the "[This system was unable to provide an English error message](https://preview.redd.it/ca8mdmw8i7xd1.png?width=888&format=png&auto=webp&s=b4593f9c56ee8e07ced1d7de4fd5365fcb4af0f4)" issue - this screenshot is of a win10 in a Portuguese-language localisation; so I downloaded pt-BR win10 iso from MS and installed in a VM.

For that specific error (`ERROR_IO_DEVICE`), the description is stored in `kernelbase.dll.mui` in windows 10 - and pt-BR localised windows does not include that resource file in English, only in pt-BR localisation. Thus the error - the system literally does not contain an English error message!

`FormatMessageW` is [documented](https://learn.microsoft.com/en-us/windows/win32/api/winbase/nf-winbase-formatmessagew), when `dwLanguageId==0`, to try getting the error message in the following order:
1. Language neutral
2. Current thread locale
3. User default locale
4. System default locale
5. English US

Therefore, this PR changes the relevant code to use `dwLanguageId==0`, while also changing the thread locale to English US inside the function, so if no English localisation is present for a certain error code description, it will first try to fall back to a localisation which is installed on the system, and only if no localised message is present for an error code, will it fall back to the default "unable to provide" string.

### DIFF
--- a/src/stdio.c
+++ b/src/stdio.c
@@ -234,10 +234,22 @@ const char *WindowsErrorString(void)
 
 	DWORD size, presize;
 	DWORD error_code, _error_code, format_error;
+	LCID locale;
 	HANDLE hModule = NULL;
 
 	error_code = GetLastError();
 	_error_code = error_code;
+	// Set thread locale to en-US when in this function.
+	// This is because kernel32!FormatMessage is documented to try the following order when dwLanguageId==0:
+	// - Language neutral
+	// - Thread locale
+	// - User default locale
+	// - System default locale
+	// - English US
+	// Some Windows localisations do not provide English MUI resources for specific error codes.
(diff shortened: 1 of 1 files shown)


pbatard:
Thanks a lot!

I had made a note to investigate this in #2576, since I saw quite a few more instances than I'd like about this issue (and my impression was that this might have been introduced with a recent change I applied with error message handling, mostly to avoid #2524), but I sure appreciate somebody else figuring that one out.

Your changes look good, but I'm still planning to test generation of errors, on a localized version of Windows, before I apply the PR.

PR #2704, opened 2025-03-13
Fix GDI leaks

ozone10:
There are still some GDI leaks, which need further investigation.

### DIFF
--- a/src/hash.c
+++ b/src/hash.c
@@ -1888,17 +1888,19 @@ INT_PTR CALLBACK HashCallback(HWND hDlg, UINT message, WPARAM wParam, LPARAM lPa
 {
 	int i, dw, dh;
 	RECT rc;
-	HFONT hFont;
+	static HFONT hFont = NULL;
 	HDC hDC;
 
 	switch (message) {
 	case WM_INITDIALOG:
 		apply_localization(IDD_HASH, hDlg);
-		hDC = GetDC(hDlg);
-		hFont = CreateFontA(-MulDiv(9, GetDeviceCaps(hDC, LOGPIXELSY), 72),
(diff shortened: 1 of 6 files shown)


ozone10:
@pbatard 
hopefully this PR should fix all GDI leaks I have found.
I'm using [GDIView](https://www.nirsoft.net/utils/gdi_handles.html) to view GDI handle count.
I've tried to open every possible rufus windows to check for leaks, but still I have not figured how to open "selection" and "list" dialogs.

pbatard:
Thanks. For the selection dialog, you need to select a Windows ISO (preferably Windows 11), then pick _Windows To Go_ under _Image option_, and click START. You will then get a selection dialog for the version as well as the Windows User Experience customization.

For the list dialog, this is a bit more tricky, as you need to trigger a disk creation failure and have an application with some kind of access to the disk. The latter is easy to do, by just having a command prompt opened on the drive, but the former is more difficult and I would say that whatever applies to the selection dialog should apply to the list dialog anyway, so if you pick up selection dialog leaks, I'll look up for similar issue for the list one.

pbatard:
> Should I squash commits to single one?

I'd prefer it, but it's up to you.

ozone10:
Ok I will check those dialogs and if there is no issue I will squash commits.

pbatard:
Thanks again for the patch.

I have now applied it, with some minor modifications (such as creating a `safe_delete_object()` macro to reduce the boilerplate on `WM_NCDESTROY`).

ozone10:
@pbatard 
you're welcome.
Unfortunately we cannot use `DeleteObject` on imagelists and icons, they use `ImageList_Destroy` and `DestroyIcon`.

pbatard:
Oops, my bad. I'll fix this in a future update.

PR #2730, opened 2025-04-21
feat: added option to use utc time

KMatuszak:
This feature adds an option to use UTC time in Windows which is helpful when dualbooting Windows with Linux in order to prevent both OSes screwing time in each other.


pbatard:
As much as I always appreciate a well crafted PR, I'm afraid I have to reject this for the reasons explained [here](https://github.com/pbatard/rufus/wiki/FAQ#user-content-Could_you_add_support_for_ltWindows_option_Xgt_in_the_Windows_User_Experience_dialog).

The WUE is purposefully limited to options that impair a streamlined installation of Windows (or that cannot be easily configured post install, such as setting the internal drives offline for Windows To Go).

Otherwise, I hope you can appreciate that everybody who wants their favourite Windows customisation option will start to send PRs (since there would be a precedent for it) and the WUE dialog will end up having hundreds of options for all the various options that are available from `unattend.xml` and that can also be configured post install.

And yes, even if I gather that booting without using UTC might have an impact to other OSes that are installed on the same machine, I am pretty confident that the number of people who dual boot Windows with other OSes and will be bothered with having to configured UTC post install is low enough not to want to add a prominent option that few people will actually have a need for.

See also: https://github.com/pbatard/rufus/wiki/FAQ#user-content-Why_cant_you_just_add_the_feature_I_want.

As such, I have to reject this PR, while still thanking you for submitting it.

PR #2779, opened 2025-07-24
Added a full disk error while saving vhd

Kazkans:
When you saving a `.vhdx` to a full disk it didn't show any error, giving the false impression that it succeeded.

I added an error so the user can see that operation didn't succeded.

The easiest way to reproduce this error is to create a new small partition and try to write a file to it.


pbatard:
Looks good, and this is definitely something Rufus should check. I'll apply your PR when I get a chance, thanks!

PR #2780, opened 2025-07-27
Fix broken dark mode on old Windows 10 builds

### LINKED ISSUE #2766: Broken window drawing on older versions of Windows 10.
Checklist
---------
- [x] I have been made aware that if my problem can be summarized as _"I've created or tried to create a media using Rufus, and ..."_, and I am not including a log, **this issue will be summarily closed**.
- [x] I clicked the 'Log' button (🗒️) or pressed <kbd>Ctrl</kbd>-<kbd>L</kbd> in Rufus, or used [DebugView](https://learn.microsoft.com/en-us/sysinternals/downloads/debugview), and copy/pasted the log into the section that says `<FULL LOG>` below.

Additionally (if applicable):
- [x] I also tried one or more of the following:
  - [x] Running Rufus on a different computer.

Issue description
-----------------
On older versions of Windows 10, the dark mode implementation seems to be rather broken. It checks for 22H2 (sounds unnecessary, wouldn't 1809 or maybe 1607 work?) and doesn't seem to fully disable dark mode when the build is too low. I've attached a screenshot of a build from master running on 1809.

Image

The problem

Image

How it looks with the minimum version lowered. Seems mostly fine. The settings button was removed by me, it appears normally on upstream.

Log
---
```
Rufus x86 v4.10.2265
Windows version: Windows 10 Enterprise LTSB x86 (Build 17763.316)
Grub versions: 0.4.6a, 2.12
Localization set to 'en-US'
Found non-USB removable device 'VMware Virtual NVMe Disk' => Eliminated
0 devices found
```

ozone10:
fix #2766

### DIFF
--- a/src/darkmode.c
+++ b/src/darkmode.c
@@ -58,7 +58,7 @@ static inline BOOL IsAtLeastWin10Build(DWORD buildNumber)
 
 static inline BOOL IsAtLeastWin10(void)
 {
-	return IsAtLeastWin10Build(WIN10_22H2);
+	return IsAtLeastWin10Build(WIN10_1809);
 }
 
 static inline BOOL IsAtLeastWin11(void)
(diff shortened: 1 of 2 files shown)


ozone10:
@Smu1zel
could you check, thx.

Smu1zel:
> @Smu1zel could you check, thx.

One second.

Smu1zel:
> > @Smu1zel could you check, thx.
> 
> One second.

Turns out I deleted my LTSC 2019 VM. Let me reinstall.

Smu1zel:
@ozone10 

image

Trying out x64 Windows this time and it looks good. Title bar is now dark as well, when it wasn't originally.

image

Running the x86 build through WOW64 correctly disables dark mode, which your code intends. This might not be what you want, though. See below.

image

Commenting out the #if wall makes dark mode work just fine on the x86 build, which makes me believe this check is unnecessary.

Your changes seem good, though I'd reevaluate whether we really need to exclude x86 users.

ozone10:
@Smu1zel 
cool, thanks for testing. The x86 is mainly intended as fallback for users if something with dark mode goes wrong, so they don't have to wait for fix. 
Currently dark mode on Windows 11 is using 3 undocumented functions. I don't expect MS to change them in future builds, but it is better to be safe and have fallback.
IMHO most Windows 10 users are using 64-bit versions of the OS, and since Windows 11 is only 64-bit, it's very likely that users will use the x64 builds of Rufus.

Of course the decision is on @pbatard.
Should dark mode be enabled for the x86 build?

pbatard:
I do want dark mode enabled for x86. I see no real reason not to have it.
But I also hear your concerns, so in the slightly modified version of the patch I will be applying, I introduce a new Rufus settings that people can add to the registry/ini file, called `DisableDarkMode` that will disable dark mode if needed.

Thanks again for working on this!

PR #2792, opened 2025-08-26
[loc] update Korean translation

VenusGirl:
Spelling corrections and additional translations


pbatard:
I'll take the typos, but I will not take the new messages because, as I believe I have explained a few times before, new messages are subject to change until I start a formal translation round (and as a matter of fact `MSG_350` did change today), and during a formal translation round, I do provide context for the messages that should help you and other translators provide a more accurate translation.

It's just too problematic to get new messages piecemeal (especially as I am pretty sure I'm going to add new messages before the next translation round), so I have to be firm that, from now on, outside of a formal translation, I will only take typos/improvements for existing messages, but I will **not** take unsolicited new message translations.

I hope you can understand my position.

PR #2803, opened 2025-09-10
Add Linux Port: GTK GUI and CLI Support for Rufus

JacobEsanders:
## Overview

This pull request introduces a comprehensive Linux port of Rufus, bringing the excellent Windows USB formatting utility experience to Linux systems.

### 🖥️ GTK Graphical Interface
- Modern GUI closely matching the original Windows Rufus interface
- Device Selection with detailed drive information and safety indicators
- Filesystem Support: FAT32, NTFS, ext2, ext3, ext4
- Volume Labeling with validation and suggestions
- Real-time Progress tracking with detailed logging
- Multi-threaded Operations keeping the UI responsive
- Safety Confirmations preventing accidental data loss

### 💻 Command-Line Interface
- Full CLI support for automation and scripting
- Batch Operations for system administrators
- Pipeline-friendly output and error handling
- All GUI features available from command line

### 🛡️ Safety Features
- Removable Device Detection - only operates on safe targets
- Comprehensive Warnings before destructive operations
- Device Information display (size, type, USB/removable status)
- Input Validation and error handling throughout

### Windows API Compatibility Layer
Clean abstraction layer (linux_compat.c/h) that translates Windows operations to Linux equivalents:
- File Operations: CreateFile → open() with proper flags
- Device Control: DeviceIoControl → Linux ioctl() operations
- Device Enumeration: Windows APIs → Linux sysfs parsing
- Error Handling: Windows error codes → Linux errno translation

### Modern Linux Integration
- Uses Linux sysfs (/sys/block) for device detection
- Integrates with system formatting tools (mkfs.*)
- Follows XDG standards for desktop integration
- Proper privilege handling and user feedback

## 📦 Files Added

- src/linux_compat.c/h - Windows API compatibility layer
- src/main.c - CLI interface implementation
- src/rufus_gui.c - GTK GUI implementation
- README.linux.md - Linux-specific documentation
- rufus-linux.desktop - Desktop integration
- Makefile - Linux build system

This addresses the long-standing community request for Linux support, bringing the reliable Rufus experience to Linux users everywhere.

**This contribution brings Rufus to Linux users everywhere! 🐧**

### DIFF
--- a/README.linux.md
+++ b/README.linux.md
@@ -1,0 +1,131 @@
+# Rufus Linux Port
+
+This branch contains a Linux port of the popular Rufus USB formatting utility, featuring both graphical (GTK) and command-line interfaces.
+
+## About This Port
+
+This Linux port is a community-driven effort to bring the excellent Rufus experience to Linux users. The port maintains the same philosophy, interface design, and safety features as the original Windows version while integrating properly with Linux systems.
+
+## Key Features
+
+- **🖥️ GTK GUI**: Modern interface closely matching the original Windows Rufus
+- **💻 CLI Interface**: Full command-line support for automation and scripting
+- **🛡️ Safety First**: Multiple protection layers prevent accidental data loss
+- **🐧 Linux Native**: Uses standard Linux tools and follows best practices
+- **📦 Cross-Distribution**: Works on Ubuntu, Fedora, Arch, and more
+
+## Installation
+
+### Dependencies
+
+#### Ubuntu/Debian
+```bash
+sudo apt-get install build-essential pkg-config libgtk-3-dev dosfstools ntfs-3g e2fsprogs
+```
+
+#### Fedora/Red Hat
+```bash
+sudo dnf install gcc pkg-config gtk3-devel dosfstools ntfs-3g e2fsprogs
+```
+
+#### Arch Linux
+```bash
+sudo pacman -S gcc pkg-config gtk3 dosfstools ntfs-3g e2fsprogs
+```
+
+### Building
+
+```bash
+# Clone the repository
+git clone https://github.com/JacobEsanders/rufus-linux.git
+cd rufus-linux
+git checkout linux-port
+
+# Build
+make
+
+# Launch GUI (requires root for device access)
+sudo ./rufus-linux-gui
+
+# Or use CLI
+sudo ./rufus-linux --list
+sudo ./rufus-linux --device /dev/sdX --filesystem fat32 --label "MYUSB"
+```
+
+## How to Use
+
+### Step 1: Connect Your USB Drive
+Insert your USB drive into an available USB port on your computer.
+
+### Step 2: Launch Rufus Linux
+Start the application using one of the methods in the Installation section.
+
+### Step 3: Choose Your Options
+
+#### Using the GUI:
+1. **Select your USB device** from the dropdown menu (carefully verify it's the correct one!)
+2. **Choose a filesystem format** (FAT32 is most compatible)
+3. **Enter a volume label** (optional)
+4. **Select an ISO file** if you want to create a bootable drive
+5. **Click 'START'** to begin the process
+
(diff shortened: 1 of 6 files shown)


pbatard:
I hate to have to do this but I'm going to have to be honest here.

As it stands, this effort appears to be "merely" (but this is not meant to be dismissive of the work carried out) a standalone application that provides a `dd` and `mkfs` equivalent/wrapper, along with a GUI.

It however fails to provide the basic must-have features such as ISO extraction support (ISO mode) through libcdio and WIM support for Windows User Experience patching, which, IMO, are two of the **basic** features any application that claims to be "Rufus for this or that OS" must have. If a Rufus "clone" cannot create a UEFI:NTFS bootable Windows installation media, with the WUE options (such as TPM bypass), then, I'm afraid that, as far as I am concerned, I cannot endorse it as a Rufus clone **because this is what users of your OS will be expecting first and foremost from Rufus**. And that's not counting the many other features such as DOS bootable drive creation, zeroing, bad blocks check (with fake drive detection), UEFI bootloader revocation validation, USB flash drive vs HDD/SDD detection (so as not to list external hard drive by defaults, even if they are removable),  and quite a few of the features that I do consider essential to any Rufus clone (even if I very much understand that one has to start somewhere, but again, seeing that none of the original code from Rufus is being reused, I can't help but have serious doubts as to these being provided to users). And I am quite willing to let Windows To Go (difficult to do on OSes where you can't invoke `bcdboot`/`bcdedit`), or ISO downloads (relies on Fido which is a PowerShell script) or saving to VHDX/FFU (relies on Windows APIs, though saving to dd/uncompressed VHD should actually be quite easy -- you could even reuse the old Rufus code from before we used the Windows APIs) fall by the wayside for a non Windows port. But if the starting point is just `dd` + `mkfs` with no reuse of the original code (especially after I went the trouble of switching to wimlib, which should make Windows bootable creation a lot easier on non Windows OSes), I don't really see much of any incentive with championing this PR, even as a starting point towards a Linux port.

As such, since, IMO, this is really a separate Linux application in disguise, that just happens to present a UI that looks somewhat similar to Rufus, but that fails the basic design of any Rufus port, which should reuse as much as the existing code as possible, I don't think I can see myself merging this into the Rufus codebase. It simply doesn't make sense to me to merge what I can only see as two really separate projects, where one of the projects attempted to copy part of the UI from the other.

This is not to say that I don't agree that there is a demand for Rufus for Linux or Mac (as a matter of fact, I am keeping my eyes open for projects like https://github.com/maciejwaloszczyk/Remus or https://github.com/THETECHYSASQUATCH/rufus which are quite a few in a somewhat growing line of people attempting to develop a port of Rufus for non-Windows platforms), but I do feel like I am perhaps well placed to have some idea of what such an effort will require in order to provide some of the core features that I have listed above, and therefore, I'd much prefer seeing something that does integrate some of these core features, preferably by reusing the many cross-platform libraries that Rufus already uses (such as ms-sys, libcdio, e2fsprogs, wimlib, syslinux and hopefully Bled since it's really busybox in disguise) before I jump into integrating any Linux port as a starting point.

Now, I will not close this PR right away, as I definitely want to leave you a chance to provide a counterpoint to what I am stating, if you feel like I'm missing the mark. But, at this stage, and despite an effort that I truly do appreciate, I don't really see myself integrating this work into Rufus. As a matter of fact, I'd much rather if people like yourself, and the ones behind the projects I have mentioned above, talked to one another, instead of each starting their own separate project, to collaborate on a single effort, because, again, I sure wouldn't mind a Linux or OS-X port of Rufus... as long as the basic core features that users will expect from a Rufus clone are provided.

JacobEsanders:
Thanks for the feedback I'll definitely work on implementing these changes

JacobEsanders:
Also I think Combining these multiple scattered projects from multiple devs into a singular more polished port is a great idea

PR #2808, opened 2025-09-21
[loc] Add Georgian translation

EkaterinePapava:
Hello,
Adding Georgian translation for Rufus.

BR, Ekaterine

### DIFF
--- a/res/loc/po/ka-GE.po
+++ b/res/loc/po/ka-GE.po
@@ -1,0 +1,2010 @@
+msgid ""
+msgstr ""
+"Project-Id-Version: 4.5\n"
+"Report-Msgid-Bugs-To: pete@akeo.ie\n"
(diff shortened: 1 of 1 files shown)


pbatard:
I'm afraid I have to reject this translation [for the reasons explained in the FAQ](https://github.com/pbatard/rufus/wiki/FAQ#what-about-the-languages-that-arent-listed-above).

Georgian is not in the [list of languages for which there exists a full localisation of Windows 11](https://www.microsoft.com/en-us/windows/windows-11-specifications#table6) and as such, because I deliberately have to limit the languages supported by Rufus to the ones Windows is fully localized into, I can not add a Georgian localisation to Rufus.

I'm really sorry to see the time you spent on this being wasted, but I just don't have the scope to manage a hundred translations or more, which is what Rufus would end up supporting if I made an exception for you (because in the spirit of fairness I should then also allow every other language).

PR #2873, opened 2025-12-08
Improved Microsoft Dev Drive detection

MartinKuschnik:
The `IsMsDevDrive` function has been refactored to use the `FSCTL_QUERY_PERSISTENT_VOLUME_STATE` control code for detecting Microsoft Dev Drives. This replaces the previous approach of analyzing drive layout which was not always correct.

Old:
old_impl

New:
new_impl


pbatard:
Can I ask you where exactly you got the information that `FSCTL_QUERY_PERSISTENT_VOLUME_STATE` is associated with dev drives, because I cannot see that information online.

Especially, [the Microsoft documentation for this flag](https://learn.microsoft.com/en-us/windows-hardware/drivers/ifs/fsctl-query-persistent-volume-state) shows that it has existed since Windows 7, and Windows 7 obviously did not have dev drives...

My worry is that, by filtering out all drives with `FSCTL_QUERY_PERSISTENT_VOLUME_STATE` we are going to filter out more than dev drives, which is not what we want.

If the issue was that some non dev drives were listed by Rufus, and that these drives had the `FSCTL_QUERY_PERSISTENT_VOLUME_STATE` flag (but they still matched the layout that Rufus attempts to detects), then the better solution is to keep the existing code **and** add a new test for the `FSCTL_QUERY_PERSISTENT_VOLUME_STATE` flag to filter out these.

Otherwise, I fear that, in some months' time, someone will come with a new PR because they found that the approach of only checking `FSCTL_QUERY_PERSISTENT_VOLUME_STATE` applies to more than dev drives, and we'll be back to square one.

MartinKuschnik:
Of course you may ask where I got the information from. I was looking for a similar solution for another project and looked at the current implementation from Rufus. In doing so, I noticed that the current implementation is more of a guessing approach and doesn't work for Dev Drives with different storage sizes.

After a longer search, I found the solution shown in other GitHub repositories. With the knowledge about the flag, you can also find something in the [WDK](https://github.com/MicrosoftDocs/windows-driver-docs-ddi/blob/7d3debf065b8e6090a9b2032e4f34a7e106a610d/wdk-ddi-src/content/ntifs/ns-ntifs-_file_fs_persistent_volume_information.md?plain=1#L69):

```markdown
### -field VolumeFlags

The persistent state settings for a file system volume. This value is a bitwise OR combination of the following.

| Value | Meaning |
| ----- | ------- |
...
| PERSISTENT_VOLUME_STATE_DEV_VOLUME 0x00002000 | Indicates that this volume is formatted as a developer volume. This can be used by the file system and other system components to enable optimizations that don't require an administrator to trust the volume on a given machine. Valid starting with Windows 11, version 22H2 September Update. |
```

And since I consider the current solution to be incorrect or insufficient, because anyone can create a VHD with this format (`MSFT_Volume::Format(DevDrive=True)`), I wanted to suggest this solution to you.

pbatard:
Oh, it looks like I browsed your PR too fast and mistook `FSCTL_QUERY_PERSISTENT_VOLUME_STATE` (the ioctl function) for  `PERSISTENT_VOLUME_STATE_DEV_VOLUME` (the actual flag).

Obviously the flag is designed to detect dev drives, in which case, this PR works for me, and can indeed greatly simplify the code.

Sorry about the noise.

pbatard:
Applied. Thanks again!

PR #2904, opened 2026-01-30
Update Korean

VenusGirl:
Delete non-translator....
This is my sole translation.


pbatard:
I'm sorry but regardless of the changes being made, there was a Korean translation carried out by these 2 persons, and even if you choose to revert all their changes, I believe it is in bad form to want to remove the fact that they contributed.

Let's put it this other way: If someone comes and chooses to re-do your Korean translation from scratch, do you believe that I should completely erase the fact that you did contribute a Korean translation?

As such, I will **not** remove the names of previous translators, no matter how much or how little they contributed.

VenusGirl:
Then I will change the order of translators to the most recent.

pbatard:
No. The order will remain chronological, as it is with every other translation. This is **not** a competition.

PR #2905, opened 2026-01-30
Update Korean Fix

VenusGirl:
Delete non-translator...

This is my sole translation.


pbatard:
Rejected, for the reasons detailed on the other issue.

PR #2937, opened 2026-03-11
Japanese translation update and fix

reindex-ot:
I understand your policy regarding unsolicited PRs, but I've created a Draft PR just for your reference.
I noticed that the current Japanese translation contains several inaccuracies and inconsistencies.
My update also ensures the terminology aligns better with the standard Japanese Windows environment.


pbatard:
Thanks for this submission.

This however is a rather massive change, so I don't think I can apply it without consulting with the existing Japanese translator. Also, you edited the `.loc` directly, whereas you are supposed to use Pollock, as described in https://github.com/pbatard/rufus/wiki/Localization#editing-an-existing-homepage-translation.

There is a translation update round coming up for the upcoming 4.14 release, so I will point the current translator to your changes for review.

pbatard:
Oh, there is also something very strange about the source you used, as your patch shows:
```
 ######################################################################
 l "ja-JP" "Japanese (日本語)" 0x0411
-v 1.0.22
+v 4.13
 b "en-US"
```

Whereas the `.loc` version for Japanese has been `v4.5` for some time.

This leads me to think that you worked with a very old version of the `.loc` in the first place, which will make it very difficult to use anything from this PR.

So, whereas I will still try to reference this PR to the current Japanese translator, and let them decide what they want to do with it, I will also close it, as I don't think it is relevant to the current localization content.

PR #2954, opened 2026-04-17
[loc] update Lithuanian translation to latest

arvaidasre:
## Summary
- update the lt-LT localization block to 4.14
- add the missing Lithuanian translations for the latest WUE, DBX and UEFI-related messages from ChangeLog.txt
- polish several existing Lithuanian UI strings to use more natural, plain-language wording

## Testing
- not run (translation-only change)


pbatard:
Thanks for this. Are you the same person who also sent me an e-mail one hour before you created this PR?

If not, I'm going to have to first push the changes proposed by the first person, and then ask if you want to review them, as I'll have to pick the first person who contacted me for precedence (and I apologize about having to do that, as it'll just be a bad coincidence that 2 different persons contacted me within one hour of each other 7 days after I made the request).

arvaidasre:
I'm not a different person, everything is fine, I can check new words, I've made it more Lithuanian there.

pbatard:
Ah okay, in that case I will pick your update from the PR, and apply that one.

arvaidasre:
Let me clarify, I'm a different person, not the same.

pbatard:
Ah... That makes sense, because the names I got were really different. In that case, as indicated above, I will push that first person's changes and then, if you feel like some updates are needed, you can create a new PR (which, in the spirit of fairness I will ask that other translator to have a look at, to make sure they are okay with it).

pbatard:
Thanks for understanding. I have now pushed the other update, which you can also see at 74efb16c9dace064fdd32666d2ab9d9c4efc4906. Feel free to propose changes on top of it as you see fit.

arvaidasre:
I'm sorry, but he didn't translate anything there. I still see it in English where it should be in Lithuanian. Everything is Lithuanianized completely as it should be for me.

pbatard:
I'm afraid I cannot see your inline comments with regards to the translation (GitHub issue, I tried multiple browser, they just don't show anything. But I do get the e-mail notifications from it), so I cannot comment on what you highlighted specifically. It did test the new messages of the WUE dialog prior to pushing the request though, and did not see anything that stroke me as odd or untranslated (especially as my instructions are to keep terms like QoL in English).

Again, if you submit a PR with what you think should be improved, I will see the changes you propose and I can submit them for review.

Sorry about the extra work, but, again, the other person contacted me before you so I have to give them precedence in the spirit of fairness.

arvaidasre:
everything okay. Good day Sir

Issue #2080, opened 2022-11-07
[INTERNAL] List of rufus-3.22 enhancements and fixes


pbatard:
﻿﻿**This topic is flagged as INTERNAL and is therefore NOT open for comments.**

- [x] [3db3bf527fe32c5f6715377e38204be9a2cd7d07] [fac433a0fef3434ddb0c44bd53e0476f7bad04b0] [3281f6b97effa7b8af77164b61e48b608cd64b87] Figure out the best way to have translators add translations to the Windows Store listings &rarr; _Finally... And people still think that anything that has to do with translation is a 5 mins job! Oh and partner.microsoft.com still has a bug where you can't upload screenshots that use the same name, even if they reside in separate directories... **which they have now fixed**_
- [x] Start a new translation round &rarr; _Finally!!!_
  - [x] [76ec89032c549fb810f5488addb7a72121261203] Make sure `MSG_199` and `MSG_294` get a translation (through DeepL or whatever) regardless of whether other messages are translated as we will need those moving forward. 
- [x] Missing translations:
  * Arabic
  * Croatian
  * Danish
  * Japanese
  * Lithuanian
  * Malay
  * Norwegian
  * Slovenian
  * Thai
  * Vietnamese
- [x] [21ac145a4b65fe09e8d88954a7c66d20b0cefe0b] Cancelling during `WinMountImage()` / `WimUnmountImage()` may leave the user with opened `boot.wim`'s that generate an error on subsequent runs. While these can be removed with `dism /Get-MountedWimInfo` + `dism /Unmount-Wim /discard/Mountdir:C:\Users\###\AppData\Local\Temp\Ruf###.tmp`, do we want to add our own unmount?
- [x] [6ebe13eb31ef3e6f3d5b9efe8f826f161693435f] Remove Fido support from Windows 7 platforms (since we're going to retire it and it **is** gonna break)
- [x] [8d04e5f1f47146a78f8b4d25f686cb60ccc89d95] Update the build `.yml`'s to convert the use of the `set-output` command per [this](https://github.blog/changelog/2022-10-11-github-actions-deprecating-save-state-and-set-output-commands/)
[...]

PR #2961, opened 2026-04-30
Update Malay (Bahasa Melayu) translation

good-work-great-work1:
I realise issue #2080 is closed — submitting anyway in case this can be included
in a future update. Please let me know if this cannot be accepted.

### DIFF
--- a/res/loc/po/ms-MY.po
+++ b/res/loc/po/ms-MY.po
@@ -1,9 +1,9 @@
 msgid ""
 msgstr ""
-"Project-Id-Version: 3.22\n"
+"Project-Id-Version: 4.14\n"
 "Report-Msgid-Bugs-To: pete@akeo.ie\n"
-"POT-Creation-Date: 2023-04-25 13:26+0100\n"
-"PO-Revision-Date: 2023-04-25 13:33+0100\n"
+"POT-Creation-Date: 2026-04-30 21:19+0300\n"
+"PO-Revision-Date: 2026-04-30 21:44+0300\n"
 "Last-Translator: \n"
 "Language-Team: \n"
 "Language: ms_MY\n"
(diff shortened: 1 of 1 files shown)


pbatard:
Hey, thanks a lot for answering the call to provide a much needed Malay translation update.
However, your proposal is missing all new new messages that have been added since 3.22, so I would really need you to follow the directions from https://github.com/pbatard/rufus/wiki/Localization#Editing_an_existing_translation as you would be able to both edit the existing messages as well as see all the new messages that need translation in the convenience of the [Poedit](https://poedit.com/) well known translator tool.

Is that something you think you can do? The Malay translation is in desperate need of a proper update for all the new strings, and just updating the existing content simply won't cut it I'm afraid...

good-work-great-work1:
Okay, I'll take a look now.

good-work-great-work1:
@pbatard Pete, I'm on Linux. Is there any way to do this without that application? Or could you give me proper instructions for installing the .exe? I can't find them anywhere.

pbatard:
> I'm on Linux. Is there any way to do this without that application?

I'm afraid I cannot support Linux, especially as I want people to be able to test their changes in real-time so that they can tweak them, which is a feature Pollock.exe is designed to offer, but which of course cannot work on Linux, since Rufus is Windows-only in the first place.

All in all, it makes no sense for me to invest time in Linux users, especially in the limited context of translation work.

I'm afraid that I have to require that any translator who wishes to contribute to Rufus must be able to access a Windows platform when doing so.

If this is not something you can do, then that's okay. I will just wait for another Malay speaker to come along, who has access to Windows.

good-work-great-work1:
@pbatard Okay, I'll sort it out on my own. Thanks.

pbatard:
I appreciate that. But again, if it's too inconvenient for you, don't worry. Regardless of the outcome, I already appreciate very much that you volunteered to help!

good-work-great-work1:
@pbatard I've done everything. Will you accept the work?

good-work-great-work1:
It's hard on Linux, but possible.

pbatard:
Great! Thanks for this. Sadly you're a few hours too late for the 4.14 release, but I'll make sure your update is added to the next version of Rufus.

pbatard:
I have now applied your translation update into Rufus. Thanks again!

PR #2968, opened 2026-05-06
Fix compressed .vhd.zst image writes

jaredhagel2:
## Summary
Fix failures when writing Windows disk images stored as compressed fixed VHDs such as `.vhd.zst`.

## Problem
When writing a compressed Windows image to a CFast card, Rufus could fail shortly after pressing START with a compressed-image write error.

The immediate write failure was caused by sector-sized writes reaching the physical device from decompressor buffers that were not sector-aligned. Some physical media accept sector-sized writes only when the source buffer is aligned as well, so the write path could fail even though the decompression itself was fine.

There was also a second issue specific to compressed fixed VHD payloads. Rufus could treat a `.vhd.zst` image like a generic compressed raw image and fall back to the compressed archive size instead of the decompressed disk payload size. That could cause the write path to stop at the archive size instead of the real end of the fixed VHD payload.

## Fix
This change:

- routes unaligned full-sector decompressor output through an aligned bounce buffer before writing to the target device
- keeps the existing partial-sector buffering behavior for compressed-image writes
- derives the decompressed content size for zstd-backed fixed VHD payloads and uses that payload size when deciding where the real disk data ends
- avoids truncating compressed fixed-VHD writes to the compressed archive size

## Result
Compressed `.vhd.zst` images now write correctly to physical media that require aligned raw writes, and fixed-VHD payloads are written to their proper payload length instead of being cut off at the compressed archive size.

### DIFF
--- a/src/format.c
+++ b/src/format.c
@@ -70,6 +70,11 @@ badblocks_report report = { 0 };
 static float format_percent = 0.0f;
 static int task_number = 0, actual_fs_type;
 static unsigned int sec_buf_pos = 0;
+// Aligned bounce buffer state for compressed-image writes.
+static size_t sec_buf_size = 0;
+static uint64_t compressed_target_size = 0;
+static uint64_t compressed_output_pos = 0;
+static BOOL compressed_extra_data_ignored = FALSE;
 extern const int nb_steps[FS_MAX];
 extern const char* md5sum_name[2];
 extern uint32_t dur_mins, dur_secs;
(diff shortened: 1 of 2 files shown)


pbatard:
Thanks for the PR. I will review it when I have a chance, but right now it's a bit low priority.

Since this is about fixing an actual failure, I don't really mind including this, but I also wouldn't mind knowing if there exists any public `.vhd.zst` made available for download, as it sure (currently) looks like a very specific use case of Rufus for an internal usage scenario.

I kinda have to wonder why someone would use a `.vhd.zst` when a VHD is really a plain "DD" image with a 512 byte footer, and they could as well just use `.img.zst` instead (which I do assume works, as the alignment issue is really a Microsoft limitation with their VHD APIs, and you didn't patch anything from Bled, which is the library that does our decompression).

So can you please elaborate on why you have a workflow that uses `.vhd.zst` instead of `.img.zst`?

Still, if having proper support for it in Rufus can encourage folks to publish ztsd compressed VHDs in the future, and if it fixes real issues with using compressed VHDs, I don't have a major problem with adding this patch. But this PR does seem oddly specific to a specific scenario that I don't imagine many people are likely to run into...

jaredhagel2:
Thanks Pete. Something has come up and this has become lower priority for us as well, plus the workflow of us using a .vhd.zst file has likely changed. I marked this pull request as 'Draft' now because I think this pull request may need some further changes to cover all cases, but because this has become lower priority for us now I don't have time to make these further changes.

PR #2978, opened 2026-05-13
[wue] add option to enable the built-in Administrator account

corestudio-dev-ai:
Adds a new checkbox to the Windows User Experience dialog for Windows 11 ISOs that, when selected, runs 'net user Administrator /active:yes' as a FirstLogonCommand in the generated unattend.xml. The option is persisted across sessions but disabled by default, since the built-in Administrator account has no password set and grants full elevated access.


pbatard:
Has been rejected in the issue you opened for it, where you appeared to indicate that you wanted to wait for an answer before proposing the PR.

PR #2986, opened 2026-05-28
Add opt-in Mica backdrop support on Windows 11 22H2+

KoukeNeko:
## Summary

Adds an opt-in Mica system backdrop for Rufus dialogs on Windows 11 22H2+ (build 22621), reusing the existing delay-loaded `dwmapi` rather than pulling in the Windows App SDK / WinRT `MicaController`.

- **Title bar Mica** for every dialog via `DwmSetWindowAttribute` (`DWMWA_SYSTEMBACKDROP_TYPE = DWMSBT_MAINWINDOW`).
- **Client-area Mica (dark mode only)** via `DwmExtendFrameIntoClientArea` with a subclass that paints the background transparent through DWM's compositor (`BeginBufferedPaint` + `BufferedPaintSetAlpha`). Restricted to dark mode because GDI's dark text is unreadable on the chroma-keyed light-mode backdrop.
- **Transparent toolbars** so they blend with the backdrop instead of painting a solid strip.
- **Flicker-free expand/collapse** of the advanced sections via a redraw freeze around the resize.
- **Opt-in only**: disabled by default. Toggle via the new checkbox in the Update Policy & Settings dialog (`Ctrl+Alt+M` shortcut also works). Takes effect after a restart.

The feature is fully gated: on anything below Windows 11 22H2, or with the
setting off, behaviour is unchanged.

image
image
image

## Test plan

- [x] Win11 22H2+ dark mode: enable Mica, confirm title bar + client area show the backdrop
- [x] Win11 22H2+ light mode: enable Mica, confirm only the title bar gets Mica and all labels stay readable
- [x] Expand/collapse advanced device & format sections: no flicker, log window does not pop up
- [x] Toggle the setting checkbox and `Ctrl+Alt+M`: both write the registry key and apply after restart
- [x] Win10 / Win11 21H2: confirm no behavioural change (feature unavailable, checkbox disabled)
- [x] Mica off (default): dialogs render exactly as before

### DIFF
--- a/src/darkmode.c
+++ b/src/darkmode.c
@@ -209,9 +209,15 @@ void SetMicaBackdrop(HWND hWnd)
 	enum DWM_SYSTEMBACKDROP_TYPE backdrop = DWMSBT_MAINWINDOW;
 	MARGINS margins = { -1, -1, -1, -1 };
 
-	if (!IsAtLeastWin11_22H2())
+	if (!IsAtLeastWin11_22H2() || ReadSetting32(SETTING_MICA) != 1)
 		return;
 	DwmSetWindowAttribute(hWnd, DWMWA_SYSTEMBACKDROP_TYPE, &backdrop, sizeof(backdrop));
+	// Light-mode GDI text is naturally dark; ClearType edges drift toward
+	// pure black which DWM's chroma-key then erases, making labels vanish.
+	// Skip the client-area extension in light mode so only the title bar
+	// gets Mica - matching how most first-party Win11 apps behave.
+	if (!is_darkmode_enabled)
+		return;
 	DwmExtendFrameIntoClientArea(hWnd, &margins);
 	if (!GetWindowSubclass(hWnd, MicaBackgroundSubclass, (UINT_PTR)MicaBackgroundSubclassID, NULL))
 		SetWindowSubclass(hWnd, MicaBackgroundSubclass, (UINT_PTR)MicaBackgroundSubclassID, 0);
--- a/src/darkmode.c
+++ b/src/darkmode.c
@@ -149,17 +149,60 @@ void SetDarkTheme(HWND hWnd)
 	SetWindowTheme(hWnd, is_darkmode_enabled ? L"DarkMode_Explorer" : NULL, NULL);
 }
 
-// Mica system backdrop (Win11 22H2+). DWM blends desktop wallpaper through the
-// non-client area; the client area only shows Mica where the window does not
-// paint over it, so Rufus's solid dark/light brushes will limit the effect to
-// the title bar. Safe to call on any Windows version.
+// Extend Mica into the client area via DWM frame extension + chroma-key.
+// Within the extended frame, pure black (RGB 0,0,0) is composed as transparent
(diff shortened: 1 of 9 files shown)


pbatard:
Thanks for the submission. I appreciate this PR and the time you spent on it, but I fail to see a real pain point that it would address as I am not seeing flicker or toolbar blending issues on current Windows 11 platforms, and I haven't seen reports of it either.

Also, if I were to add it, it will **not** be as an option, as most users won't have the slightest clue what Mica is and what benefit they might get from it and it's purely cosmetic. We don't even have a UI option to disable or force dark mode/light mode, so I don't see why we should have a UI option for Mica. Please be mindful that Rufus is not Winamp, which means that we don't aim at having skins or UI candy so that users can relish in what they are looking at while creating a bootable drive, and it looks to me like the main point of Mica is that it'll allow users to use different shades of dark colors for their backdrop, or make them semi-transparent, which, again, I fail to see as a "must have" feature.

So, with regards to having this as an option, if Mica is that great and brings us demonstrable benefits, then it should be applied unconditionally on any platform that supports it.

Also I'm pretty sure that, since we still pop system dialogs here and there that I don't believe you patched for Mica, it will make our UI look very inconsistent, with the main dialog supporting Mica, but most of the popup dialogs not supporting it, and therefore looking very jarring.

So, to be honest, unless you can demonstrate that the Mica feature can address real UI pain-points that users experience with Rufus, as was the case for dark mode, this feature looks more trouble than it's worth, as it'd mean that (if optional) we'd now have TWO different paths to maintain/test for UI elements, and I'd rather concentrate on non UI matters than making things look pretty for whatever new UI scheme Microsoft invariably tries to push forward every couple of years or so, before it promptly abandons it for a completely different UI scheme. IMO, time would be better spent switching Rufus to full WinUI using XAML islands, than trying to pick bits and pieces that expect WinUI, just so that we can kind of fool users into thinking that they are dealing with a modern WinUI application.

So can you please explain what this brings to the table, what existing UI problems in Rufus you can demonstrate it addresses, and why it would make sense for Rufus users to have it (in which case, again, there is no reason why it should be optional)?

KoukeNeko:
Indeed, as someone whose first real experience of owning and understanding a computer was with Windows 10, I may naturally feel that UI is an important part of UX. Later on, one of the reasons I moved away from Windows and switched to macOS was the UI, so I suppose I may be more sensitive to this than the typical Rufus user.

That said, perhaps it is exactly Rufus’ lightweight, no-nonsense approach that makes it so easy to use. Adding Mica may look nice, but I understand that it may not be the best choice for Rufus.

Also, regarding the flickering I observed, it was probably caused by a rendering issue on the remote lab computer I was using, rather than Rufus itself. So I do not want to overstate that as an actual issue with the application.

My main frustration is probably less about Rufus itself, and more about the fact that native Win32 UI on Windows often feels rather dated.

I fully understand your reasoning, and I appreciate you taking the time to explain it. I’ll close this pull request.

Thank you again for your feedback, and thank you for maintaining Rufus. Maybe one day, when I return to Windows, I will appreciate Rufus even more for staying simple, focused, and reliable.

PR #3007, opened 2026-06-29
restore classic right click context menu as qol

Fredolx:
Restore the classic right click context menu as part of the QoL enchancements.

image


pbatard:
Not a bad idea, as this is indeed a QoL regression Microsoft inflicted to Windows users. This won't be part of 4.15 (too late in the release process) but I'll consider it for the next version.

pbatard:
Applied, thanks.

I also disabled auto-installation of manufacturer's app when plugging new hardware, as part of this WUE patch, mostly due to [this documented very questionable practice by hardware manufacturers](https://www.youtube.com/watch?v=Q9uefFYe6bM).

lextra2:
> 
> I also disabled auto-installation of manufacturer's app when plugging new hardware

@pbatard 
Could you also add a checkbox to prevent Windows from automatically downloading drivers?

To achieve this, you'd want to change `SearchOrderConfig` under `HKEY_LOCAL_MACHINE\SOFTWARE\Microsoft\Windows\CurrentVersion\DriverSearching` from `1` to `0`

Please note that this checkbox should also set `PreventDeviceMetadataFromNetwork` to `0`, or at least I think that this is a requirement.

pbatard:
@lextra2, I will only disable the automated driver installation of drivers by setting `PreventDeviceMetadataFromNetwork` to `0`.
I have no plans (and I see no reason) to go further than this, as this change was motivated by the video I linked.

PR #3014, opened 2026-07-14
[ui] add completion notification and log elapsed time

d06i:


pbatard:
No. This has been requested before, but it's completely **duplicate/redundant** information, and you have to be very conscious that when it comes to UI preferences, something that you may find desirable can actually be annoying for the vast majority of users.

Rufus uses the standard Windows cues for completion (full progress bar, return to active buttons, taskabar progress icon reset, etc) and displays the time elapsed prominently. There is no real need to add a super "in your face" display of the same cues, as it's going to quickly start to annoy users who know what they are doing (_"Why are you telling me you're done, when I can see plainly that you are done?"_).

pbatard:
Also Rufus is not a benchmarking tool, so keeping track of how long an operation took in the log is not something I see any need for. If you want to benchmark how fast your drive is, use an actual benchmarking application rather than rely on an estimate that is highly dependent on what image you are writing and the specific drive you used which won't tell you anything more than _"Oh yeah, I spend x minutes and y seconds writing that drive"_.

PR #3026, opened 2026-07-28
Update it-IT.po

lucamosca1:
Fixed translation errors and internet set lowercase. Feel free to discard if not needed


pbatard:
Applied. Thanks!

PR #3032, opened 2026-08-17
cregex: fix signed-char OOB/UB in char classes and 0-byte alloc on empty pattern

SilentSobs:
Found via fuzzing the cregex module.

- Fixes undefined behavior from signed-char indexing into character-class lookup tables (a char with the high bit set becomes a negative index, causing an out-of-bounds read).
- Avoids a 0-byte allocation when compiling an empty pattern.

### DIFF
--- a/src/cregex.h
+++ b/src/cregex.h
@@ -77,12 +77,17 @@ typedef char cregex_char_class[(UCHAR_MAX + CHAR_BIT - 1) / CHAR_BIT];
 static inline int cregex_char_class_contains(const cregex_char_class klass,
                                              int ch)
 {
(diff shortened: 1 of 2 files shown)


pbatard:
Thanks for the Pull Request. I think you also want to create a similar PR in https://github.com/jserv/cregex since this is the original code we are reusing so it has the same issue.

pbatard:
Applied. Thanks again for reporting this both here and to the upstream repo.

pbatard:
Closed in d3fd56c000b614bfa09e1e5d799f3867417057f4. I messed the reference there and I'm too lazy to fix the patch... 😉

PR #3033, opened 2026-08-17
cregex: fix heap OOB read past pattern buffer in parse_char_class()

SilentSobs:
Found via fuzzing the cregex module.

Fixes a heap out-of-bounds read where the "empty range" check in a character class compared a possibly negative, sign-extended char against the next byte, letting the parser advance past the end of the pattern buffer. Also rejects a dangling backslash-escape at end-of-pattern the same way an unescaped NUL already is.

Also submitted upstream to the original library: jserv/cregex#2


pbatard:
Applied. Thanks again for reporting this both here and to the upstream repo.

PR #3034, opened 2026-08-18
Show filtered USB drives as hidden hints

scorpi023:
## Summary

- Keep USB HDD/SSD and large-card candidates hidden by the default hard-drive filter visible as informational device rows suffixed with `(hidden)`.
- Show Rufus's native information dialog when a hidden row is activated, explaining how to enable `List USB Hard Drives` under Advanced drive properties.
- Preserve the existing explicit opt-in and all hard safety exclusions.

## Safety

- Informational rows use sentinel item data and are appended after real drives, preserving the combo-index-to-drive-array mapping.
- A hidden selection restores the previous real target (or no selection) before the dropdown closes and before the modal opens.
- Hidden rows cannot reach `PopulateProperties`, process search, or formatting paths.
- Microsoft Dev Drives, explicitly filtered drives, Windows Sandbox/Bitdefender VHDs, the system disk, and the disk hosting Rufus remain excluded.

## Validation

- Debug x64 build with MSVC v143 override: passed.
- Release x64 build with MSVC v143 override: passed.
- `git diff --check`: passed.

### DIFF
--- a/res/loc/ChangeLog.txt
+++ b/res/loc/ChangeLog.txt
@@ -11,6 +11,9 @@ o v?.??
   - *NEW*      MSG_372 "I WILL ensure that all disks, except the one I want to install Windows on, are disconnected."
   - *NEW*      MSG_373 "I WILL ensure that I don't leave this media plugged on a PC I do not plan to erase."
   - *NEW*      MSG_374 "I AGREE that any data loss resulting from the use of this option lies entirely with me."
+  // These messages identify drives hidden by the default USB hard-drive filter and explain how to enable them.
+  - *NEW*      MSG_375 "%s — %s (hidden)"
+  - *NEW*      MSG_376 "To use this drive, enable the 'List USB Hard Drives' option in Advanced drive properties."
 
 o v4.14 (2026.04.30)
   // NB: I used Google translate to update the following message, so please check for accuracy!
--- a/res/loc/rufus.loc
+++ b/res/loc/rufus.loc
@@ -641,6 +641,8 @@ t MSG_371 "If you use this option, please make sure to disconnect every disk fro
 t MSG_372 "I WILL ensure that all disks, except the one I want to install Windows on, are disconnected."
 t MSG_373 "I WILL ensure that I don't leave this media plugged on a PC I do not plan to erase."
 t MSG_374 "I AGREE that any data loss resulting from the use of this option lies entirely with me."
+t MSG_375 "%s — %s (hidden)"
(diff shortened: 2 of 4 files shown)


pbatard:
I'm most likely going to have to shoot this down. IMO this will both make it more confusing for users and defeat the goal of **not** making it too easy to access drives that users, who are unlikely to be careful with said drives, may end up erasing by mistake...

The underlying design of Rufus in that respect is to preserve data that should be preserved, and not as you assert, to make it easier for first time users to figure out why Rufus does not show all of their drives and how they can (partially) remedy that. So the listing of large drives/HDDs is deliberately made obtuse and hard to guess for (what we think is) a very good reason. If we wanted intuitiveness/ease of use over data preservation, we'd have altered the UI to do something along the lines of what you suggest some time ago, but we kept away from doing that precisely because we assert that it will leave the door open for reckless behaviour and angry Rufus users who happened to reformat an important backup drive by mistake, because me made it a bit _too easy_ for them to list that drive...

Therefore, I am not planning to approve this PR.

scorpi023:
This will still not directly select the drive and/or enable for flashing it. All it adds is the information to enable if the drive is disabled. I ran into this issue when I tried using a 128GB USB drive and had to search why it isn't visible. The functionality of having to select the 'Enable USB Hard Drives' option to actually flash it is still preserved. Does it help?

pbatard:
> I ran into this issue when I tried using a 128GB USB drive and had to search why it isn't visible.

Then please create an issue for this or e-mail if you believe it should be listed by default. I regularly adjust the HDD vs UFD filter of Rufus precisely to list drives that Rufus might be a bit too eager to remove and if you look at the log you would have seen a line like:

```
Found UAS (USB 3.1) device 'PNY USB 3.2.2 FD UAS Device' (154B:1007)
Device eliminated because it was detected as a Hard Drive or SSD (score 3 > 0)
If this device is not a Hard Drive or SSD, please e-mail the author of this application
NOTE: You can enable the listing of Hard Drives under 'advanced drive properties'
0 devices found
```

> All it adds is the information to enable if the drive is disabled.

Which, again, we purposefully don't want to make too easy to figure out by inexperienced users who **will** end up formatting the wrong drive because of it. There's more than enough _"How can I recover data from this drive I formatted by mistake using Rufus"_ on public forums and social media sites to make me **not** want to make it easier for people to figure out how to list drives that are not listed by default. And as I said elsewhere, I'd rather inconvenience a few people, while preserving data of some others, than satisfy the first few at the cost of the latter losing some data they didn't want to lose.

scorpi023:
Opened https://github.com/pbatard/rufus/issues/3035 to address this.

